PR #99, opened 2023-08-28
feat: eth_getTransactionLog

ratankaliani:
## Features
- Add eth_getTransactionLog

## Refactor
- Update order of variables for eth_getStorageAt to reflect natural order `blockHash`, `address`, `storageKey`
- Update offsets to be more explicit

### DIFF
--- a/plonky2x/src/frontend/eth/storage/generators/storage.rs
+++ b/plonky2x/src/frontend/eth/storage/generators/storage.rs
@@ -14,7 +14,7 @@ use sha2::Digest;
 use tokio::runtime::Runtime;
 
 use crate::frontend::builder::CircuitBuilder;
-use crate::frontend::eth::storage::vars::{EthLogVariable, EthLog};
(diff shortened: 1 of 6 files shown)


plonky2x/src/frontend/eth/storage/generators/storage.rs
```diff
@@ -118,3 +120,114 @@ impl<F: RichField + Extendable<D>, const D: usize> SimpleGenerator<F, D>
...
+                    .await
+                    .expect("Failed to call get_transaction_receipt")
+            })
+            .expect("No transaction receipt found");
+
+        let log = &result.logs[self.log_index as usize];
+        let value = EthLog {
+            address: log.address,
+            topics: [log.topics[0], log.topics[1], log.topics[2]],
+            data_hash: ethers::types::H256::from_slice(sha2::Sha256::digest(&log.data).as_ref()),
```
puma314, line 186: very small nit: let's just import `H256` and `Sha256` directly from the crates instead of having these long names

puma314 (approved):

PR #100, opened 2023-08-28
feat(plonky2x): eth_getAccount

ratankaliani:
## Feat
- Add generator for `eth_getAccount`

## Misc
- Refactor order of functions in `eth/storage/builder.rs`, `eth/storage/generators/storage.rs`, `eth/storage/vars/storage.rs`


plonky2x/src/frontend/eth/storage/vars/storage.rs
```diff
@@ -62,10 +62,11 @@ impl CircuitVariable for EthProofVariable {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub struct EthAccount {
     pub balance: U256,
     pub code_hash: H256,
+    // Update to U64 once we have U64 variables
```
puma314, line 69: I think now we have `U64` variables, so we can just use that here

puma314 (commented):
1 small nit, other than that looks great

plonky2x/src/frontend/eth/storage/vars/storage.rs
```diff
@@ -192,10 +190,9 @@ impl CircuitVariable for EthLogVariable {
     }
 
     fn from_variables(variables: &[Variable]) -> Self {
-        let mut var_buffer = VariableBuffer::new(variables);
```
ratankaliani, line 195: `VariableBuffer` doesn't seem to be supported by `circuit.test_default_serializers`

PR #109, opened 2023-08-30
feat(plonky2x): ssz merkle proofs

ameya-deshmukh:


plonky2x/src/frontend/eth/beacon/builder.rs
```diff
@@ -40,7 +45,47 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+    ) -> Bytes32Variable {
+        let mut hasher = leaf; // Initialize the hasher with the leaf node
+        for i in 0..depth { // Iterate over each level of the Merkle tree
+            let (first, second) = if (index >> i) & 1 == 1 { 
+                // Determine the order of hashing based on the index
+                (branch[i].as_bytes(), hasher.as_bytes())
+            } else {
+                (hasher.as_bytes(), branch[i].as_bytes())
+            };
+            let data: [ByteVariable; 64] = first.iter()
```
puma314, line 75: nit: there must be a cleaner way to do this? maybe using `.concat`?

plonky2x/src/frontend/eth/beacon/builder.rs
```diff
@@ -40,7 +45,47 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
         self.add_simple_generator(&generator);
         generator.validator
     }
+
+    /// SHA256 implementation for builder 
+        pub fn sha(&mut self, input: &[ByteVariable]) -> Bytes32Variable {
```
puma314, line 50: I think we should add this to `frontend/hash/sha/sha256` instead of here

plonky2x/src/frontend/eth/beacon/builder.rs
```diff
@@ -40,7 +45,47 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+            let input_bool: Vec<BoolTarget> = input.iter().flat_map(|byte| byte.0.iter().cloned().map(|bool_variable| BoolTarget::new_unsafe(bool_variable.0 .0))).collect();
```
puma314, line 51: nit: I think we should have a method on `ByteVariable` that just directly returns a slice of BoolTargets so that we can reuse the logic?

plonky2x/src/frontend/eth/beacon/builder.rs
```diff
@@ -40,7 +45,47 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+            let hash_bool = sha256::sha256::<F,D>(&mut self.api, &input_bool);
+            let hash_bytes_vec = hash_bool.chunks(8).map(|chunk| ByteVariable(array![i => BoolVariable::from(chunk[i].target); 8])).collect::<Vec<_>>();
```
puma314, line 53: same nit here, we can add another method to ByteVariable to instantiate it from bool targets so that we can reuse the logic

puma314 (commented):
This looks awesome!! Nice work on figuring out our APIs, which aren't the easiest to figure out and also not well documented!

Some small nits and then also would be great to have a test showing this working! For an example, check out an example of how we do testing here:

https://github.com/succinctlabs/succinctx/blob/main/plonky2x/src/frontend/uint/uint256.rs#L162

plonky2x/src/frontend/eth/beacon/builder.rs
```diff
@@ -40,7 +45,47 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+            hash_bytes_array.copy_from_slice(&hash_bytes_vec);
+            Bytes32Variable(BytesVariable(hash_bytes_array))
+        }
+    
+    /// SSZ Merkle Proof 
+    pub fn ssz_restore_merkle_root(
+        &mut self,
+        leaf: Bytes32Variable,
+        index: u64,
+        depth: usize,
```
jtguibas, line 64: I think we don't actually need depth as an argument. It can use https://github.com/succinctlabs/telepathy/blob/main/contracts/src/libraries/SimpleSerialize.sol#L30-L48 to automatically deduce the depth.

plonky2x/src/frontend/eth/beacon/builder.rs
```diff
@@ -40,7 +45,36 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
         self.add_simple_generator(&generator);
         generator.validator
     }
+
+    /// SSZ Merkle Proof
```
jtguibas, line 49: nit: use full sentences for function comments. Maybe something like

`Computes the expected merkle root given a leaf, branch, and deterministic index.`

plonky2x/src/frontend/eth/beacon/builder.rs
```diff
@@ -40,7 +45,36 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+    pub fn ssz_restore_merkle_root(
+        &mut self,
+        leaf: Bytes32Variable,
+        index: u64,
+        branch: Vec<Bytes32Variable>,
+    ) -> Bytes32Variable {
+        assert!(2u64.pow(branch.len() as u32 + 1) > index);
+        let mut hasher = leaf; // Initialize the hasher with the leaf node
+        for i in 0..branch.len() { // Iterate over each level of the Merkle tree
+            let (first, second) = if (index >> i) & 1 == 1 {
```
jtguibas, line 59: nit:  comments are overly used, can remove all of them. the code can speak for itself

plonky2x/src/frontend/hash/sha/sha256.rs
```diff
@@ -276,6 +279,16 @@ pub fn sha256<F: RichField + Extendable<D>, const D: usize>(
     process_sha256(builder, &msg_input)
 }
 
+/// SHA256 implementation for builder 
+pub fn sha_for_builder<F: RichField + Extendable<D>, const D: usize>(builder: &mut Plonky2xCircuitBuilder<F,D>, input: &[ByteVariable]) -> Bytes32Variable {
+    let input_bool: Vec<BoolTarget> = input.iter().flat_map(|byte| byte.to_bool_targets().to_vec()).collect();
+    let hash_bool = sha256::<F,D>(&mut builder.api, &input_bool);
+    let hash_bytes_vec = hash_bool.chunks(8).map(|chunk| ByteVariable(array![i => BoolVariable::from(chunk[i].target); 8])).collect::<Vec<_>>();
```
jtguibas, line 286: can you instead implement it as a function ontop of builder?

impl<F, D> CircuitBuilder {
 pub fn sha256(...)
}

plonky2x/src/frontend/vars/byte.rs
```diff
@@ -192,6 +193,12 @@ impl<F: RichField + Extendable<D>, const D: usize> RotateRight<F, D, usize> for
 impl<F: RichField + Extendable<D>, const D: usize> Zero<F, D> for ByteVariable {
     fn zero(builder: &mut CircuitBuilder<F, D>) -> Self {
         ByteVariable(array![_ => builder.constant(false); 8])
+    }    
+}
+
+impl ByteVariable {
+    pub fn to_bool_targets(&self) -> [BoolTarget; 8] {
```
jtguibas, line 200: instead of `to`, could you use `as_bool_targets(&self)`?

jtguibas (approved):

PR #111, opened 2023-08-31
feat: add mux to builder

ctian1:


plonky2x/src/frontend/builder/mux.rs
```diff
@@ -0,0 +1,55 @@
+use plonky2::field::extension::Extendable;
+use plonky2::hash::hash_types::RichField;
+
+use crate::prelude::{CircuitBuilder, CircuitVariable, Variable};
+
+trait MuxBuilder {
```
jtguibas, line 6: trait is unecessary, you can impl directly on builder

PR #123, opened 2023-09-03
feat: mux

justcode740:
anyone could help figure out what's wrong with current impl?

### DIFF
--- a/plonky2x/src/frontend/builder/mux.rs
+++ b/plonky2x/src/frontend/builder/mux.rs
@@ -10,21 +10,24 @@ trait MuxBuilder {
 impl<F: RichField + Extendable<D>, const D: usize> MuxBuilder for CircuitBuilder<F, D> {
     fn select_index<T: CircuitVariable>(&mut self, selector: Variable, inputs: &[T]) -> T {
         let num_var_in_t = inputs[0].variables().len();
-        let input_len = inputs.len();
-        let mut res = (0..num_var_in_t).map(|_| self.init::<Variable>()).collect::<Vec<_>>();
         let api = &mut self.api;
+        let mut final_vars = Vec::new();
 
-        for i in 0..input_len {
-            let target_i = api.constant(F::from_canonical_usize(i));
-            let whether_select = api.is_equal(target_i, selector.0);
-
-            let vars = inputs[i].variables();
-            for j in 0..num_var_in_t {
-                res[j] = Variable(api.select(whether_select, vars[j].0, res[j].0));
+        for i in 0..num_var_in_t {
+            // All the variables in the `i`-th position for each element of input
+            let mut pos_vars = Vec::new(); 
+            
+            for j in 0..inputs.len() {
+                pos_vars.push(inputs[j].variables()[i].0);
             }
-        }
+            // Pad the length of pos_vars to the nearest power of 2, random_access constrain len of selected vec to be power of 2
+            let padded_len = pos_vars.len().next_power_of_two();
+            pos_vars.resize_with(padded_len, Default::default);
 
-        T::from_variables(&res)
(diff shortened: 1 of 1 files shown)


plonky2x/src/frontend/builder/mux.rs
```diff
@@ -9,7 +9,22 @@ trait MuxBuilder {
...
+        let mut res = (0..num_var_in_t).map(|_| self.init::<Variable>()).collect::<Vec<_>>();
+        let api = &mut self.api;
+
+        for i in 0..input_len {
+            let target_i = api.constant(F::from_canonical_usize(i));
+            let whether_select = api.is_equal(target_i, selector.0);
+
+            let vars = inputs[i].variables();
+            for j in 0..num_var_in_t {
+                res[j] = Variable(api.select(whether_select, vars[j].0, res[j].0));
```
puma314, line 23: Because here we are selecting between many different variables and `api.select` is only meant for selecting between 2 variables, you should use the `random_access` gate in the `plonky2` api:

https://github.com/mir-protocol/plonky2/blob/main/plonky2/src/gadgets/random_access.rs#L14

If you plug this into this PR, I think you should be all set!

plonky2x/src/frontend/builder/mux.rs
```diff
@@ -9,7 +9,22 @@ trait MuxBuilder {
 
 impl<F: RichField + Extendable<D>, const D: usize> MuxBuilder for CircuitBuilder<F, D> {
     fn select_index<T: CircuitVariable>(&mut self, selector: Variable, inputs: &[T]) -> T {
-        todo!()
+        let num_var_in_t = inputs[0].variables().len();
+        let input_len = inputs.len();
+        let mut res = (0..num_var_in_t).map(|_| self.init::<Variable>()).collect::<Vec<_>>();
```
puma314, line 14: Here you're basically just allocating the variables for the result. I think it's better to not do this and instead just allocate an empty `Vec` and `push` to the vec, which will be easier after you use the `random_access` gadget that I wrote the comment about below

puma314 (commented):
Nice start! Just use the `random_access` gadget (I wrote a comment about it) and you should be all set

plonky2x/src/frontend/builder/mux.rs
justcode740: yeah, think that's how it's used here? not fully get how `random_access` can be plugged here since it applies to vec<target> but not vec<vec<target>>

puma314: Here is the pseudocode:

```
final_vars = Vec::new()
for i in 0..num_var_it_t {
   pos_vars = Vec::new() // all the variables in the `i`-th position for each element of input
   for j in 0..inputs.len() {
      pos_vars.push(inputs[j].variables()[i])
   }
   final_vars.push(api.random_access(selector, pos_vars));
}

return T::from_variables(final_vars)
```

let me know if the above doesn't make sense

justcode740:
alternative impl with select is here: https://github.com/succinctlabs/succinctx/pull/125, unsure best way to compare performance now

plonky2x/src/frontend/builder/mux.rs
```diff
@@ -9,7 +9,25 @@ trait MuxBuilder {
 
 impl<F: RichField + Extendable<D>, const D: usize> MuxBuilder for CircuitBuilder<F, D> {
     fn select_index<T: CircuitVariable>(&mut self, selector: Variable, inputs: &[T]) -> T {
-        todo!()
+        let num_var_in_t = inputs[0].variables().len();
```
puma314, line 12: This is great to have this implementation in addition to your other one. Do you want to add this implementation to the other PR as another method `select_index_with_gate` and then benchmark/compare the 2 methods? I'm personally curious to see which one will come out ahead.

PR #125, opened 2023-09-05
feat: mux with select

justcode740:


plonky2x/src/frontend/builder/mux.rs
```diff
@@ -0,0 +1,71 @@
+use plonky2::field::extension::Extendable;
+use plonky2::hash::hash_types::RichField;
+
+use crate::prelude::{CircuitBuilder, CircuitVariable, Variable};
+
+trait MuxBuilder {
+    fn select_index<T: CircuitVariable>(&mut self, selector: Variable, inputs: &[T]) -> T;
+}
+
+impl<F: RichField + Extendable<D>, const D: usize> MuxBuilder for CircuitBuilder<F, D> {
```
puma314, line 10: nit: here I think you can just remove the trait `MuxBuilder` and implement it directly on `CircuitBuilder`--our mistake for including that in the original template

plonky2x/src/frontend/builder/mux.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn select_index<T: CircuitVariable>(&mut self, selector: Variable, inputs: &[T]) -> T {
+        let num_var_in_t = inputs[0].variables().len();
```
puma314, line 12: nit: can you add comments as to what these variables mean?

plonky2x/src/frontend/builder/mux.rs
```diff
@@ -0,0 +1,71 @@
...
+        let input_len = inputs.len();
+
+        let mut res = inputs[0].variables();
+        let api = &mut self.api;
+
+        for i in 0..input_len {
+            let target_i = api.constant(F::from_canonical_usize(i));
```
puma314, line 19: nit: here, we always prefer to use the `builder` instead of `api` where a builder method exists. Thus, here I would do `self.constant::<Variable>(F::from_canonical_usize(i))`.

Same with below with `builder.is_equal` and pass in 2 variables (there should be a method on the builder that allows you to compare `equal` on 2 CircuitVariables).

Where the builder method doesn't exist (like `api.select`), then feel free to use the api directly.

puma314 (commented):
Awesome work and great test. To test the performance, you can use something like this below to initialize the plonky2 logger, which will print out the number of gates and rows in the circuit. The less gates the better :)

```
let mut builder_logger = env_logger::Builder::from_default_env();
builder_logger.format_timestamp(None);        builder_logger.filter_level(log::LevelFilter::Trace);
builder_logger.try_init()?;
```

plonky2x/src/frontend/builder/mux.rs
justcode740: yeah, that's also what feel a bit weird for me : )

plonky2x/src/frontend/builder/mux.rs
justcode740: actually what's the purpose of this thin wrapper, since it leads to some unnecessary oh, e.g. unwrap and wrap, but i adjust to follow the convention.

justcode740:
> Awesome work and great test. To test the performance, you can use something like this below to initialize the plonky2 logger, which will print out the number of gates and rows in the circuit. The less gates the better :)
> 
> ```
> let mut builder_logger = env_logger::Builder::from_default_env();
> builder_logger.format_timestamp(None);        builder_logger.filter_level(log::LevelFilter::Trace);
> builder_logger.try_init()?;
> ```

for smaller value <100, random_access lighter, for larger value, random_access would overflow

PR #126, opened 2023-09-05
feat: add serialization and deserialization for custom gates and witness generators

tamirhemo:
Enables the option for users of Plonky2x to write their own gates and generators and serialize them.

Users who wish to do so will need to specify their gates and generators when implementing the `CircuitFunction` trait.


plonky2x/src/backend/circuit/serialization.rs
```diff
@@ -0,0 +1,423 @@
...
+            panic!("Generator type {} already registered", id);
+        }
+
+        self.0.type_ids.push(id.clone());
+        self.0.index.insert(id, self.0.current_index);
+        self.0.current_index += 1;
+    }
+
+    /// Registers a new simple witness generator with the given id.
+    pub fn register_simple_generator<SG: SimpleGenerator<F, D>>(&mut self, id: String) {
```
jtguibas, line 172: Rename register_simple_generator to register_simple to make the code below more readable?

plonky2x/src/backend/circuit/serialization.rs
```diff
@@ -0,0 +1,423 @@
...
+    }
+}
+
+impl<F: RichField + Extendable<D>, const D: usize> WitnessGeneratorRegistry<F, D> {
+    /// Creates a new registry with all the default generators that are used in a Plonky2x circuit.
+    pub fn new<C: GenericConfig<D, F = F> + 'static>() -> Self
+    where
+        C::Hasher: AlgebraicHasher<F>,
+    {
+        let mut registry = Self(SerializationRegistry::new());
```
jtguibas, line 288: Instead of registry, maybe just do let mut r to make the code more readabale

jtguibas (approved):

PR #127, opened 2023-09-05
feat(plonky2x): Added `ArrayVariable`

puma314:


plonky2x/src/frontend/vars/array.rs
```diff
@@ -0,0 +1,101 @@
...
+            elements: value.into_iter().map(|x| V::constant(builder, x)).collect(),
+        }
+    }
+
+    fn variables(&self) -> Vec<Variable> {
+        self.elements.iter().flat_map(|x| x.variables()).collect()
+    }
+
+    fn from_variables(variables: &[Variable]) -> Self {
+        assert_eq!(variables.len(), N);
```
tamirhemo, line 57: I think this should be
```
assert_eq!(variables.len(), N * V::nb_elements())
```

plonky2x/src/frontend/vars/array.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    fn init<F: RichField + Extendable<D>, const D: usize>(
+        builder: &mut CircuitBuilder<F, D>,
+    ) -> Self {
+        Self {
+            elements: (0..N).map(|_| V::init(builder)).collect(),
+        }
+    }
+
+    fn constant<F: RichField + Extendable<D>, const D: usize>(
```
tamirhemo, line 43: Should constant accept a slice?

plonky2x/src/frontend/vars/array.rs
puma314: yup thanks for the catch, fixed

plonky2x/src/frontend/vars/array.rs
puma314: I think it can't because of the `ValueType`, constant has to accept `Self::ValueType<F>,` which in this case is a `vec`

tamirhemo (approved):

tamirhemo (approved):

jtguibas (approved):

PR #129, opened 2023-09-06
feat(plonky2x): hint witness generation

tamirhemo:
Creates a simple generator from a fn. 

Users can register hint fn on a witness registry using the `register_hint()` method.


plonky2x/src/frontend/eth/beacon/builder.rs
```diff
@@ -217,7 +217,7 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
             withdrawals,
             idx,
         );
-        self.add_simple_generator(&generator);
+        self.add_simple_generator(generator.clone());
```
puma314, line 220: Why did you change all of this?

plonky2x/src/backend/circuit/serialization.rs
```diff
@@ -200,6 +201,24 @@ impl<F: RichField + Extendable<D>, const D: usize> WitnessGeneratorRegistry<F, D
     pub fn register_simple<SG: SimpleGenerator<F, D>>(&mut self, id: String) {
         self.register::<SimpleGeneratorAdapter<F, SG, D>>(id)
     }
+
+    pub fn register_hint(&mut self, hint_fn: fn(&mut ValueStream<F, D>, &mut ValueStream<F, D>)) {
```
jtguibas, line 205: Should we have ReadableValueStream and WriteableValueStream? Right now it looks like you could read from the second stream.

plonky2x/src/frontend/vars/buffer.rs
```diff
@@ -1,27 +1,173 @@
...
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Stream<T> {
+    data: Vec<T>,
+    position: usize,
+}
+
+impl<T> Stream<T> {
```
puma314, line 55: If `Stream` has nothing to do with `Vars`, which it seems like it doesn't would prefer to maybe put it in `vars/utils` to make it clear that itself is not a var? 

Also I'm surprised there's no existing Rust type for this...

plonky2x/src/backend/circuit/serialization.rs
puma314: I agree with this

plonky2x/src/frontend/generator/simple.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn run_once(&self, witness: &PartitionWitness<F>, out_buffer: &mut GeneratedValues<F>) {
+        self.generator.run(witness, out_buffer);
+    }
+
+    fn serialize(&self, dst: &mut Vec<u8>, _common_data: &CommonCircuitData<F, D>) -> IoResult<()> {
+        let bytes = bincode::serialize(self).map_err(|_| IoError)?;
+        dst.write_all(&bytes)
+    }
+
+    fn deserialize(src: &mut Buffer, _common_data: &CommonCircuitData<F, D>) -> IoResult<Self>
```
puma314, line 61: It would be great to have a test here as an example of what adding a `Generator` is like?

plonky2x/src/frontend/generator/hint.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::backend::circuit::serialization::{GateRegistry, WitnessGeneratorRegistry};
+    use crate::prelude::{ByteVariable, *};
+
+    fn plus_one<F: RichField + Extendable<D>, const D: usize>(
+        input_stream: &mut ValueStream<F, D>,
```
puma314, line 142: Can you put all this stuff in the prelude so that our users have an easy time importing it?

plonky2x/src/frontend/generator/hint.rs
```diff
@@ -0,0 +1,185 @@
...
+        circuit.verify(&proof, &input, &output);
+
+        // Read output.
+        let byte_plus_one = output.read::<ByteVariable>();
+        assert_eq!(byte_plus_one, 6u8);
+
+        // Test the serialization
+        let gate_serializer = GateRegistry::new();
+        let mut generator_serializer = WitnessGeneratorRegistry::new::<PoseidonGoldilocksConfig>();
+        generator_serializer.register_hint(plus_one);
```
puma314, line 182: Do we have to `register_hint` on the `circuit` anywhere?

plonky2x/src/frontend/vars/buffer.rs
```diff
@@ -1,27 +1,173 @@
...
+use serde::{Deserialize, Serialize};
+
 use super::{CircuitVariable, Variable};
-pub struct VariableBuffer<'a> {
-    variables: &'a [Variable],
-    position: usize,
+use crate::prelude::CircuitBuilder;
+
+#[derive(Debug, Clone)]
+pub struct ValueStream<F, const D: usize>(Stream<F>);
```
puma314, line 12: can we add comments on what each of these things are meant to be?

plonky2x/src/frontend/eth/beacon/builder.rs
jtguibas: You shouldn't use the generator after you've added it

plonky2x/src/frontend/vars/buffer.rs
```diff
@@ -1,27 +1,173 @@
+use core::marker::PhantomData;
```
jtguibas, line 1: rename this file to stream?

plonky2x/src/frontend/vars/buffer.rs
jtguibas: I feel like it shouldn't even be in `vars`, it should be in `utils` because it's used for ValueStream.

plonky2x/src/frontend/vars/buffer.rs
```diff
@@ -1,27 +1,173 @@
...
+        variables
+    }
+    pub fn read<V: CircuitVariable>(&self, builder: &mut CircuitBuilder<F, D>) -> V {
+        let variables = self.read_exact(builder, V::nb_elements());
+        V::from_variables(&variables)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Stream<T> {
```
jtguibas, line 50: If we have ValueStream, can we also have VariableStream instead of Stream<Variable>?

plonky2x/src/frontend/eth/storage/vars/storage.rs
```diff
@@ -198,7 +198,7 @@ impl CircuitVariable for EthLogVariable {
     }
 
     fn from_variables(variables: &[Variable]) -> Self {
-        let mut var_buffer = VariableBuffer::new(variables);
+        let mut var_buffer = Stream::<Variable>::new(variables.to_vec());
```
jtguibas, line 201: Use VariableStream instead?

plonky2x/src/frontend/generator/hint.rs
tamirhemo: No, but users will have do return their own `GeneratorRegistry` when they implement CircuitFunction.

plonky2x/src/backend/circuit/serialization.rs
tamirhemo: I named them ValueReader and ValueWriter

plonky2x/src/frontend/generator/simple.rs
tamirhemo: I can add an example without variables in it, but until we implement serde::Serialize and Deserialize on variable it's kind of blocked, so maybe this should be a separate PR?

jtguibas (approved):

PR #130, opened 2023-09-06
wip bench

justcode740:


plonky2x/src/frontend/builder/mux.rs
```diff
@@ -0,0 +1,186 @@
+use plonky2::field::extension::Extendable;
+use plonky2::hash::hash_types::RichField;
+use plonky2::iop::target::BoolTarget;
+
+use crate::prelude::{CircuitBuilder, CircuitVariable, Variable};
+
+impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
+    pub fn select_index_with_select<T: CircuitVariable>(
```
jtguibas, line 8: use V instead of T

plonky2x/src/frontend/builder/mux.rs
```diff
@@ -0,0 +1,186 @@
...
+    ) -> T {
+        // Serialized variables for the circuit variable.
+        let num_var_in_t = inputs[0].variables().len();
+        let input_len = inputs.len();
+
+        let mut res = inputs[0].variables();
+
+        for i in 0..input_len {
+            let target_i = self.constant::<Variable>(F::from_canonical_usize(i));
+            let diff = self.sub(target_i, selector);
```
jtguibas, line 21: This is a bit inefficient in terms of rows. 


A better way to implement it would be as an inner product. In particular, using builder.api.add_many can sum a bunch of Targets in one row. There might also be a method for an inner product, not sure.

plonky2x/benches/mux_bench.rs
```diff
@@ -0,0 +1,37 @@
...
+    group.sample_size(10);
+
+    type F = GoldilocksField;
+    const D: usize = 2;
+    let mut builder = CircuitBuilder::<F, D>::new();
+    let selector = builder.constant::<Variable>(F::from_canonical_usize(42)); // Example selector value
+    let inputs: Vec<u64> = (1..=74851).collect();
+
+    let input_variables = inputs
+            .iter()
```
jtguibas, line 20: This benchmark is somewhat unnecessary to compare the two implementations, you can measure how expensive it is the run the circuit by printing the number of gates.

I think there is a method like builder.api.nb_gates()

jtguibas: As a result, I think we can remove this file

It's also not a fair comparison due to the way Plonky2 pads the plonk table to the next power of two

plonky2x/src/frontend/builder/mux.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        let output = builder.select_index(index_one, &input_variables[..]);
+
+        builder.assert_is_equal(output, input_variables[0]);
+
+        let circuit = builder.build::<C>();
+        let pw = PartialWitness::new();
+        let proof = circuit.data.prove(pw).unwrap();
+        circuit.data.verify(proof).unwrap();
+        // This verification should fail
```
jtguibas, line 183: nit: remove?

plonky2x/src/frontend/builder/mux.rs
justcode740: i push a manual inner product is it what u suggest? run with vec size 10 but not really improved
image

plonky2x/benches/mux_bench.rs
justcode740: this issue arise when i try to add gate print for diff impl for mux but need to back and forth, and also feel it's not very systematic.

why this is not fair? actually even with this number, random_access is fast, but the issue is current impl of select_with_random_access overflow with lot of inputs, do u mind help take a look?
https://github.com/succinctlabs/succinctx/pull/130/commits/b239731e2ab07a65be83df0d32e5cb83fa4cb038#diff-ba6371572665a5385874a1d7bf91cead2351f50f015206f1e33cbc2643a16036R128

puma314:
Closing this, as we changed around a few things around so just took the code from here and implemented it there

PR #135, opened 2023-09-08
feat(Plonky2x): derive CircuitVariable

tamirhemo:
Adds the ability to derive `CircuitVariable` for structs. This will generate another type for the value. 

```rust
         #[derive(Debug, Clone, CircuitVariable)]
        struct Point<V, U> {
            pub x: V,
            y: U,
            z: Variable,
        }
```
By default, the corresponding value will be called #typenameValue.

Attributes in this macro:

- value_name: users can choose their own name for the value type. For example:
```rust
         #[derive(Debug, Clone, CircuitVariable)]
         #[value_name(MyPoint)]
        struct Point<V, U> {
            pub x: V,
            y: U,
            z: Variable,
        }
```
will generate a value struct named MyPoint.

- [value_derive]
Users can specify traits they want to derive on the value type. For example:
```rust
         #[derive(Debug, Clone, CircuitVariable)]
         #[value_derive(PartialEq, Eq)]
        struct MyStruct {
            x: Variable,
            y: Variable,
        }
```
Will derive PartialEq, Eq on MyStructValue. This value_derive feature still doesn't work well with generic bounds (even constants). It will throw a panic message in rust analyzer.


ratankaliani:
hype!!!

ratankaliani:
bullish

jtguibas (approved):

PR #147, opened 2023-09-12
feat: Initial starter template for RLC trick for subarray

puma314:
This is an initial starter template for TODOs for implementing the random linear combination (RLC) trick for checking subarray equality. 

Additionally, here is an explanatory document that goes into more detail about this technique: https://hackmd.io/@succinctlabs/ByYXnBKA3


smsunarto:
ok

PR #153, opened 2023-09-12
feat(plonky2x): tendermint merkle tree

ratankaliani:
## Features
- Tendermint Merkle Tree functionality, conforming to the specification for the [CometBFT Merkle Tree](https://docs.cometbft.com/main/spec/core/encoding#merkle-trees).
- Fixed Length
   - Invoke `compute_root_from_leaves`, `NB_LEAVES` should be the next power of 2 after `NB_LEAVES_ENABLED`.
- Variable Length
   - If leaves are of variable length, directly invoke the `get_root_from_hashed_leaves` function once leaves are hashed.


kevjue (approved):
LGTM

PR #157, opened 2023-09-13
feat: integrate gnark plonky2 wrapper

ctian1:
* all changes from tamir/gnark_wrapper
* fix curta constrain code in builder.build by adding CurtaConfig to PlonkParameters
* modify plonky2x CLI to generate actual gnark verifier contract during build
  * and add `prove-wrapped` CLI command (takes in `--wrapper-path` to gnark circuit) which calls gnark circuit to wrap plonky2 proof


tamirhemo (approved):

plonky2x/src/frontend/recursion/verifier.rs
```diff
@@ -0,0 +1,5 @@
+// /// A variable encoding the verifier circuit data for a plonky2x proof.
```
jtguibas, line 1: plz delete

jtguibas (approved):

plonky2x/src/frontend/recursion/verifier.rs
tamirhemo: done

PR #171, opened 2023-09-16
feat: get rid of constants map

jtguibas:


plonky2x/src/frontend/vars/mod.rs
```diff
@@ -101,12 +101,12 @@ pub trait CircuitVariable: Debug + Clone + Sized + Sync + Send + 'static {
         value: Self::ValueType<L::Field>,
     ) -> Vec<L::Field> {
         let mut builder = CircuitBuilder::<L, D>::new();
-        let variable = builder.constant::<Self>(value);
-        let variables = variable.variables();
-        variables
-            .into_iter()
-            .map(|v| *builder.constants.get(&v).unwrap())
-            .collect()
+        let variables = builder.constant::<Self>(value).variables();
```
puma314, line 104: I think we should leave a comment here that using this method is quite expensive so it should be rare

puma314 (approved):

Issue #75, opened 2023-08-23
Use env_logger for debugging instead of println!


jtguibas:

jtguibas:
Resolved in #174.

PR #174, opened 2023-09-18
chore: use env_logger for debugging instead of println!

meekteek:
This PR addresses [issue #75] https://github.com/succinctlabs/succinctx/issues/75

Changes made:
-changed all println! statements to use env_logging info!() to be used. 
-small refactor on env_logger initialization

This is my first time doing open source contribution.
I would like feedback on these changes. Thank you!


jtguibas:
Looks good so far!

Could you please use utils::setup_logger(); instead of setup_logger directly?

and also, please go through the repo and look for all instantiations of env_logger::init and replace it with utils::setup_logger()

jtguibas:
Also, inside tests, debug! is more appropriate than info!.

jtguibas:
Also, you shouldn't call setup_logger in places besides entrypoints. That means calling it in random functions is inappropriate.

plonky2x/src/frontend/eth/rlp/builder.rs
```diff
@@ -25,6 +27,7 @@ pub fn bool_to_u32(b: bool) -> u32 {
 }
 // Note this only decodes bytes and doesn't support long strings
 pub fn rlp_decode_bytes(input: &[u8]) -> (Vec<u8>, usize) {
+    setup_logger();
```
jtguibas, line 30: don't need this

plonky2x/src/frontend/eth/rlp/builder.rs
```diff
@@ -49,13 +52,14 @@ pub fn rlp_decode_bytes(input: &[u8]) -> (Vec<u8>, usize) {
             1 + len_of_str_len + str_len,
         );
     } else {
-        println!("input {:?}", input);
+        info!("input {:?}", input);
```
jtguibas, line 55: debug better

plonky2x/src/frontend/eth/rlp/builder.rs
```diff
@@ -49,13 +52,14 @@ pub fn rlp_decode_bytes(input: &[u8]) -> (Vec<u8>, usize) {
...
         panic!("Invalid prefix rlp_decode_bytes")
     }
 }
 
 pub fn rlp_decode_list_2_or_17(input: &[u8]) -> Vec<Vec<u8>> {
-    println!("input {:?}", Bytes::from(input.to_vec()).to_string());
+    setup_logger();
```
jtguibas, line 61: same comments

meekteek:
Thanks for the prompt feedback. I have now:

changed all instances of setup_logger() with utils::setup_logger(), including instances I did not originally introduce (example \succinctx\plonky2x\src\backend\wrapper\wrap.rs had setup_logger() before any of my changes)

Looked for instances of env_logger::init to remove.

I see in plonky2x\src\frontend\hash\poseidon\poseidon256.rs there is a setup_logger(); within test_poseidon() that was already in place before. It does not have any log macros used (no debug!(), info!(), etc.). I did not remove it but updated to utils::setup_logger() in case there were plans to introduce log calls in the future.

changed all info!() macros within tests into debug!()

Some questions:
Found some test methods (ex. test_sha512_large_msg_variable in plonky2x\src\frontend\hash\sha\sha512.rs) had dbg!() call. should this also be changed to debug?

Regarding setup_logger usage: I am not entirely sure where setup_logger() is called on runtime and originally decided to include it in all methods that used the log macros. In my most current check-in I removed all instances of utils::setup_logger() in methods that were not tests that I introduced. Please let me know if I am wrong in doing that, but that's the message I got out of comments to take setup_logger() out of methods like plonky2x/src/frontend/eth/rlp/builder.rs

in files such as succinctx\plonky2x\src\backend\wrapper\wrap.rs prove() method I saw existing code where some non #[test] methods have debug!() instead of info!(). should all log calls be debug!() then or was there different rationale for this?

jtguibas:
@meekteek This is good progress. I think we could just merge this, any more work seems to me very marginal.

jtguibas (approved):

PR #179, opened 2023-09-19
feat: add array methods to builder

jtguibas:


plonky2x/src/frontend/builder/mod.rs
```diff
@@ -168,11 +169,35 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+
     /// Initializes a variable with a constant value in the circuit.
     pub fn constant<V: CircuitVariable>(&mut self, value: V::ValueType<L::Field>) -> V {
         V::constant(self, value)
     }
 
+    /// Initializes an array of variables with a constant value in the circuit.
+    pub fn constant_array<V: CircuitVariable, const N: usize>(
+        &mut self,
+        value: &[V::ValueType<L::Field>; N],
```
puma314, line 190: nit: why don't you just put in a slice here instead of requiring a fixed length slice?

puma314 (approved):

PR #182, opened 2023-09-19
feat(Plonky2x): asynchronous hint

tamirhemo:
This PR adds the ability to produce hints which don't block parts of the circuit independent of their output. 

- We have an `AsyncHint` trait:
```rust
#[async_trait]
pub trait AsyncHint<L: PlonkParameters<D>, const D: usize>:
    'static + Debug + Clone + Send + Sync + Serialize + DeserializeOwned
{
    /// required method
    async fn hint(
        &self,
        input_stream: &mut ValueStream<L, D>,
        output_stream: &mut ValueStream<L, D>,
    );
    
    // ...
```
which allows a hint to perform async code. These are used in a Circuit as follows:
```rust
let output_stream = builder.async_hint(input_stream, hint);
```

- We can also choose to run any normal hint in a non-blocking way. This is done using:
```rust
let output_stream = builder.async_hint(input_stream, Async(hint));
```

The trait is implemented as follows:
```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct EthStorageProofHint<L: PlonkParameters<D>, const D: usize> {
    chain_id: u64,
    _phantom: PhantomData<L>,
}

#[async_trait]
impl<L: PlonkParameters<D>, const D: usize> AsyncHint<L, D> for EthStorageProofHint<L, D> {
    async fn hint(
        &self,
        input_stream: &mut ValueStream<L, D>,
        output_stream: &mut ValueStream<L, D>,
    ) {
        let block_hash = input_stream.read_value::<Bytes32Variable>();
        let address = input_stream.read_value::<AddressVariable>();
        let location = input_stream.read_value::<Bytes32Variable>();

        let provider = get_provider(self.chain_id);
        let result = provider
            .get_proof(address, vec![location], Some(block_hash.into()))
            .await
            .expect("Failed to get proof");
        let value = u256_to_h256_be(result.storage_proof[0].value);
        output_stream.write_value::<Bytes32Variable>(value);
    }
}
```

Some remarks:
- In order for the associated futures to implement `Send` (which is needed for concurrency) we implement async in hint using the [async_trait](https://crates.io/crates/async-trait) crate. This has the unfortunate necessary to use the `#[async_trait]` macro every time we implement the trait. This is necessary because the current unstable async trait feature does not allow any trait bounds on the returned future (this is knows as the "Send problem"). 

- We also refactor and document the serialization folder in `circuit`.


plonky2x/src/backend/circuit/serialization/gates.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+impl<L: PlonkParameters<D>, const D: usize> GateRegistry<L, D> {
+    /// Registers a new gate.
+    pub fn register<G: AnyGate<L::Field, D>>(&mut self) {
+        let type_id = TypeId::of::<G>();
+        // let exists = self
+        //     .0
+        //     .registry
```
jtguibas, line 76: remove?

plonky2x/src/frontend/builder/mod.rs
```diff
@@ -147,18 +166,106 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+        }
     }
 
     pub fn mock_build(mut self) -> MockCircuitBuild<L, D> {
         if !self.sha256_requests.is_empty() {
             self.curta_constrain_sha256();
         }
-        let mock_circuit = self.api.mock_build();
+
+        for (index, gen_ref) in self
```
puma314, line 197: My feeling here is instead of duplicating all this code between `mock_build` and `build`, we should have a "pre_build` hook that contains all this code and then just call the hook in both `mock_build` and `build`. What do you think?

plonky2x/src/frontend/eth/beacon/builder.rs
```diff
@@ -41,18 +45,20 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
-            VALIDATORS_ROOT_GINDEX,
-        );
+        let mut input_stream = VariableStream::new();
+        input_stream.write(&block_root);
+        let hint = BeaconValidatorsHint::new(self.beacon_client.clone().unwrap());
+        let output_stream = self.async_hint(input_stream, hint);
+
+        let validators_root = output_stream.read::<Bytes32Variable>(self);
+        let proof = array![_ => output_stream.read::<Bytes32Variable>(self); DEPTH];
+        // let generator =
```
puma314, line 55: nit: renmove?

plonky2x/src/frontend/builder/mod.rs
tamirhemo: Yeah it would be nicer, I will add it

plonky2x/src/backend/circuit/witness.rs
```diff
@@ -1,43 +1,127 @@
...
-use plonky2::hash::hash_types::RichField;
-use plonky2::iop::generator::GeneratedValues;
-use plonky2::iop::target::Target;
-use plonky2::iop::witness::{PartialWitness, PartitionWitness, Witness, WitnessWrite};
+//! Witness generation for Plonky circuits.
+//!
+//! These methods are based on the [`generate_partial_witness`][1] method in `plonky2`, with the
+//! added functionality to have non-blocking witness generation for asynchronous hints and
+//! computationally expensive generators.
+//!
```
jtguibas, line 6: nit: why 2 lines?

plonky2x/src/backend/circuit/witness.rs
```diff
@@ -1,43 +1,127 @@
...
+//!
+//! [1] : https://github.com/mir-protocol/plonky2/blob/main/plonky2/src/iop/generator.rs#L19
+//!
+//!
```
jtguibas, line 10: nit: why 2 lines here?

plonky2x/src/backend/circuit/witness.rs
```diff
@@ -1,43 +1,127 @@
...
-    prover_data: &'a ProverOnlyCircuitData<F, C, D>,
-    common_data: &'a CommonCircuitData<F, D>,
-) -> Result<PartitionWitness<'a, F>, GenerateWitnessError> {
+pub fn generate_witness<'a, L: PlonkParameters<D>, const D: usize>(
+    inputs: PartialWitness<L::Field>,
+    prover_data: &'a ProverOnlyCircuitData<L::Field, L::Config, D>,
+    common_data: &'a CommonCircuitData<L::Field, D>,
+    async_generator_refs: &'a BTreeMap<usize, AsyncHintDataRef<L, D>>,
+) -> Result<PartitionWitness<'a, L::Field>> {
+    // If async hints are present, set up the a handler and initialize
```
jtguibas, line 34: break at 100columns, not arbitrarily

plonky2x/src/backend/circuit/witness.rs
```diff
@@ -1,43 +1,127 @@
...
+    // the generators with the handler's communication channel.
+    let (tx_handler_error, rx_handler_error) = oneshot::channel();
+    let async_generators = match async_generator_refs.is_empty() {
+        true => BTreeMap::new(),
+        false => {
+            let (tx, rx) = unbounded_channel();
+            // initialize the hint handler.
```
jtguibas, line 41: nit: why is this lowercase but the next comment is not? keep it standardized

plonky2x/src/backend/circuit/witness.rs
```diff
@@ -1,43 +1,127 @@
...
+    )
+}
+
+pub async fn generate_witness_async<'a, L: PlonkParameters<D>, const D: usize>(
+    inputs: PartialWitness<L::Field>,
+    prover_data: &'a ProverOnlyCircuitData<L::Field, L::Config, D>,
+    common_data: &'a CommonCircuitData<L::Field, D>,
+    async_generator_refs: &'a BTreeMap<usize, AsyncHintDataRef<L, D>>,
+) -> Result<PartitionWitness<'a, L::Field>> {
+    // If async hints are present, set up the a handler and initialize
```
jtguibas, line 76: break at 100

plonky2x/src/backend/circuit/witness.rs
```diff
@@ -1,43 +1,127 @@
...
+    // the generators with the handler's communication channel.
```
jtguibas, line 77: nit: capitalize

plonky2x/src/backend/circuit/witness.rs
```diff
@@ -1,43 +1,127 @@
...
+    let (tx_handler_error, rx_handler_error) = oneshot::channel();
+    let async_generators = match async_generator_refs.is_empty() {
+        true => BTreeMap::new(),
+        false => {
+            let (tx, rx) = unbounded_channel();
+            // initialize the hint handler.
```
jtguibas, line 83: nit capitalize

jtguibas (approved):
LGTM

PR #187, opened 2023-09-20
feat(plonky2x): implement `range` and `index` on `BytesVariable`

ratankaliani:
- Support index and slice on `BytesVariable`. Helps replace legacy `extract_hash_from_protobuf` in zkQGB and will be useful in parity between `BytesVariable` and `ArrayVariable`.


puma314 (approved):
beautiful

PR #191, opened 2023-09-21
feat: curta blake2b

kevjue:


plonky2x/src/frontend/builder/mod.rs
```diff
@@ -36,8 +38,13 @@ pub struct CircuitBuilder<L: PlonkParameters<D>, const D: usize> {
     pub(crate) hints: Vec<Box<dyn HintGenerator<L, D>>>,
     pub(crate) async_hints: Vec<AsyncHintDataRef<L, D>>,
     pub(crate) async_hints_indices: Vec<usize>,
-    pub sha256_requests: Vec<Vec<Target>>,
-    pub sha256_responses: Vec<[Target; 32]>,
+
+    // Right now we only have 2 accelerator, so just have
```
jtguibas, line 42: Please use full sentences with proper capitalization and end all comments with periods.

In addition, please fill out comments to 100 columns instead of using arbitrary line breaks.

plonky2x/src/frontend/hash/blake2/blake2b_curta.rs
```diff
@@ -0,0 +1,205 @@
...
+        builder.curta_constrain_blake2b(self);
+    }
+}
+
+impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
+    pub fn curta_blake2b_pad<const MAX_NUM_CHUNKS: usize>(
+        &mut self,
+        message: &[ByteVariable],
+    ) -> Vec<ByteVariable> {
+        // TODO: Currently, Curta does not support no-ops over BLAKE2B chunks. Until Curta BLAKE2B supports no-ops, last_chunk should always be equal to MAX_NUM_CHUNKS - 1.
```
jtguibas, line 40: nit: seems to be beyond 100 columns

plonky2x/src/frontend/hash/blake2/blake2b_curta.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    /// Executes a BLAKE2B hash on the given message.
+    pub fn curta_blake2b_variable<const MAX_NUM_CHUNKS: usize>(
+        &mut self,
+        message: &[ByteVariable],
+        message_len: U64Variable,
+    ) -> Bytes32Variable {
+        // TODO: Currently, Curta does not support no-ops over BLAKE2B chunks. Until Curta BLAKE2B supports no-ops, last_chunk should always be equal to MAX_NUM_CHUNKS - 1.
```
jtguibas, line 63: nit: seems to be beyond 100 columns

plonky2x/src/frontend/hash/sha/sha256_curta.rs
```diff
@@ -131,26 +132,54 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
-        )
+        accelerator.sha256_responses.push(digest);
+
+        let bytes: [ByteVariable; 32] =
+            digest.map(|x| convert_byte_target_to_byte_var(x, &mut self.api));
+        bytes.into()
+    }
+
+    /// Executes a SHA256 hash on the given input with a mutable accelerator.
+    /// Note: Should only be called by curta_constrain_sha256.
```
jtguibas, line 162: nit: https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text

there should be a line break after 161 to seperate the main description of a function from its details

plonky2x/src/frontend/hash/bit_operations/mod.rs
```diff
@@ -19,6 +20,37 @@ pub mod util;
 use util::{biguint_to_bits_target, bits_to_biguint_target};
 
 use crate::frontend::num::biguint::CircuitBuilderBiguint;
+use crate::prelude::{BoolVariable, ByteVariable, CircuitVariable};
+
+pub fn convert_byte_target_to_byte_var<F: RichField + Extendable<D>, const D: usize>(
```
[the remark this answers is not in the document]
puma314: Also it's worth checking if we already have them somewhere

plonky2x/src/frontend/vars/byte.rs
```diff
@@ -133,6 +134,40 @@ impl ByteVariable {
...
+            .expect("Expected 8 bits.  Should never happen");
+        let mut bool_variables: [BoolVariable; 8] = le_bits.map(|x| x.into());
+
+        // Need to reverse it to big endian
+        bool_variables.reverse();
+        ByteVariable::from_be_bits(bool_variables)
+    }
+
+    /// Creates a Target from a ByteVariable.
+    pub fn to_target<L: PlonkParameters<D>, const D: usize>(
```
puma314, line 161: nit: should we have these be `to_variable` and `from_variable` since we're trying to basically never operate with targets?

puma314 (commented):
Looks much cleaner to me! I'll let @jtguibas or @tamirhemo do the final approval, but left 1 small nit

plonky2x/src/frontend/vars/byte.rs
kevjue: I renamed `to_target` to `as_target` to be more similar to this function:  https://github.com/succinctlabs/succinctx/blob/main/plonky2x/src/frontend/vars/byte.rs#L109.

Also, I actually disagree about renaming `from_target` to `to_variable`.  It doesn't seem very Rust-like.  Ideally, I wanted to implement the From trait:

`impl From<Target> for BytesVariable`

but I needed a builder handler to do some of the bit manipulation, so the next best option seemed to be creating a `from_target` method.

puma314: I mean that instead of returning a target, it should return a `Variable` (like a wrapped `Target`) since in our world, we always try to work with `Variable`, not `Target`

kevjue: done

jtguibas (approved):

PR #196, opened 2023-09-25
feat(plonky2x): keccak256

sfyll:
Below is an edit; the first attempt was slightly too optimistic. 🙈

The circuit was implemented by combining Jump's keccak256 circuit with Kevin's work, removing shared hash components, and integrating the circuit "front-end" with the Succinct X "back-end." A few questions remain:

- Most notably, it seems we want to constrain the front-end to accept only an array of bytes (as indicated in `circuit.go`). If this is the agreed-upon standard for passing inputs into hashing circuits, this PR requires some tweaks to be compliant.

- Currently, the circuit has the following characteristics:
<div align="center">

| Operation    | num_gates | quotient_degree_factor |
|--------------|-----------|------------------------|
| keccak256(4)  | 25382     | 8                      |
| keccak256     | 6315      | 8                      |

</div>
    Given that the optimization guidelines are qualitative, it's unclear whether they meet the threshold.

- I welcome any other feedback, as this is my first time working with plonky2. Its versatility is a double-edged sword, making the learning curve quite steep.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2553,6 +2553,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "sha3",
  "tokio",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2553,7 +2553,6 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
- "sha3",
  "tokio",
 ]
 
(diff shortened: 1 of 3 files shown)


SFYLL:
Updated this PR to work with &[ByteVariable].

longer hashes is still WIP, in case someone else wants to check it.

PR #197, opened 2023-09-26
feat: get_fixed_subarray using RLC

kevjue:


plonky2x/src/frontend/vars/array.rs
```diff
@@ -162,18 +170,75 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+        array_size: Variable,
+        start_idx: Variable,
+        seed: &ArrayVariable<ByteVariable, 15>, // TODO: Seed it with 120 bits.  Need to figure out if this is secure.
+    ) -> ArrayVariable<Variable, SUB_ARRAY_SIZE> {
+        // TODO:  Need to add check that array_size is less than MAX_ARRAY_SIZE.
+        // TODO:  Need to add check that start_idx + SUB_ARRAY_SIZE is less than array_size.
+
+        let sub_array = self.init::<ArrayVariable<Variable, SUB_ARRAY_SIZE>>();
+
+        self.add_simple_generator(SubArrayExtractor::<L, D, MAX_ARRAY_SIZE, SUB_ARRAY_SIZE> {
```
puma314, line 185: I think that we should transition to using the `hint` api everywhere instead of `add_simple_generator`, soi I would change that here

plonky2x/src/frontend/vars/array.rs
```diff
@@ -162,18 +170,75 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+            array: array.clone(),
+            array_size,
+            start_idx,
+            sub_array: sub_array.clone(),
+            _marker: PhantomData,
+        });
+
+        let mut challenger = RecursiveChallenger::<L::Field, PoseidonHash, D>::new(&mut self.api);
+        for seed_chunk in seed.as_vec().chunks(5) {
```
puma314, line 194: If we change this over to `Variable`, then we won't have to do this chunking ourselves

plonky2x/src/frontend/vars/array.rs
```diff
@@ -162,18 +170,75 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+            let seed_element_bits = seed_chunk
+                .iter()
+                .flat_map(|x| x.as_bool_targets())
+                .collect_vec();
+            let seed_element = self.api.le_sum(seed_element_bits.iter());
+            challenger.observe_element(seed_element);
+        }
+
+        for _i in 0..2 {
```
puma314, line 203: Should we take this as an opportunity to get the elements in the extension field, instead of looping through twice and getting the challenges twice? We can also ask @tamirhemo for advice.

plonky2x/src/frontend/vars/array.rs
```diff
@@ -162,18 +170,75 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+
+            let mut accumulator1 = self.zero::<Variable>();
+            let mut j_target = self.zero();
+            for j in 0..MAX_ARRAY_SIZE {
+                let at_start_idx = self.is_equal(j_target, start_idx);
+                within_sub_array = within_sub_array.add(at_start_idx.variables()[0], self);
+                let at_end_idx = self.is_equal(j_target, end_idx);
+                within_sub_array = within_sub_array.sub(at_end_idx.variables()[0], self);
+
+                let mut subarray_idx = j_target.sub(start_idx, self);
```
puma314, line 222: Imo we should just have a modular function that gets an array commitment and then build this code on top of that

plonky2x/src/frontend/vars/array.rs
```diff
@@ -162,18 +170,75 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
-    ) -> ArrayVariable<Variable, SIZE> {
-        let mut res = Vec::new();
-        for i in 0..SIZE {
-            let i_variable = self.constant(L::Field::from_canonical_usize(i));
-            let index = self.add(start, i_variable);
-            res.push(self.select_array(array, index));
+        array: &ArrayVariable<Variable, MAX_ARRAY_SIZE>,
+        array_size: Variable,
+        start_idx: Variable,
+        seed: &ArrayVariable<ByteVariable, 15>, // TODO: Seed it with 120 bits.  Need to figure out if this is secure.
```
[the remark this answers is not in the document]
kevjue: Regarding using ByteVariable, I think the vast majority of the time, the seed would be a hash, so a bytesvariable type.

Regarding arbitrary size, specifying the seed length in the type will help enforce that during compile time, and having the check done then is generally better than run time.

plonky2x/src/frontend/vars/array.rs
kevjue: Do you mean a subarray commitment?

plonky2x/src/frontend/vars/array.rs
kevjue: @tamirhemo - Would appreciate your thoughts on this.

plonky2x/src/frontend/vars/array.rs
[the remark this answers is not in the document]
puma314: Don't you think that people might want to seed it with headers that are 32 bytes and instead of passing in something of length 15, it's easier for them to just pass in a `Vec`? The reason I'm kind of opposed to this `15` hardcoded here is that it's a minimum--it's not always the case that it's 15. Also have `ByteVariable` here to me seems overly restrictive, it's just any sort of hash or randomness in the circuit that suffices, right? That randomness could come from Variables as well?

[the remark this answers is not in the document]
kevjue: I think that the risk of having Variables passed in is that they could be representing something arbitrary (e.g. they could be bits or bytes).  So its not very clear on how many bits of security the seed would be after decomposing the variable type.  Also, rust's rng seed function takes in a vector of u8's (https://rust-random.github.io/rand/rand/trait.SeedableRng.html#associatedtype.Seed).

As for the 15 hard coded limit, that's fair.

plonky2x/src/frontend/vars/array.rs
kevjue: I think at this point, we should just have a follow up PR for this, since the deadline is not too far off ...

plonky2x/src/frontend/vars/array.rs
```diff
@@ -162,18 +168,94 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+        array: &ArrayVariable<Variable, ARRAY_SIZE>,
+        start_idx: Variable,
+        seed: &[ByteVariable],
+    ) -> ArrayVariable<Variable, SUB_ARRAY_SIZE> {
+        // TODO:  Need to add check that array_size is less than MAX_ARRAY_SIZE.
+        // TODO:  Need to add check that start_idx + SUB_ARRAY_SIZE is less than array_size.
+        const MIN_SEED_BITS: usize = 120; // TODO: Seed it with 120 bits.  Need to figure out if this is enough bits of security.
+
+        let mut input_stream = VariableStream::new();
+        for array_element in array.as_vec().iter() {
```
puma314, line 180: You can just do `input_stream.write(&array)` and it will work  :)

plonky2x/src/frontend/vars/array.rs
```diff
@@ -162,18 +168,94 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+        input_stream.write(&start_idx);
+
+        let hint = SubArrayExtractorHint {
+            array_size: ARRAY_SIZE,
+            sub_array_size: SUB_ARRAY_SIZE,
+        };
+        let output_stream = self.hint(input_stream, hint);
+
+        let mut sub_array_element = Vec::new();
+        for _i in 0..SUB_ARRAY_SIZE {
```
puma314, line 192: you can do `output_stream.read::<ArrayVariable<Variable, SUB_ARRAY_SIZE>>()` and it will work :)

plonky2x/src/frontend/vars/array.rs
```diff
@@ -162,18 +168,92 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+                .collect_vec();
+            let seed_element = self.api.le_sum(seed_element_bits.iter());
+            seed_bit_len += seed_element_bits.len();
+            seed_targets.push(seed_element);
         }
-        ArrayVariable::new(res)
+
+        assert!(seed_bit_len >= MIN_SEED_BITS);
+
+        challenger.observe_elements(seed_targets.as_slice());
```
puma314, line 215: Why can't we just do `challenger.observe_elements(seed.to_vec().flat_map(|x| x.as_targets()).collect()` there seems to be no need to convert it into elements and add additional constraints because the total amount of randomness will be the same whether it's in bit form or byte form?

kevjue: Ya, I agree that's cleaner.

I think packing the seed into golidlocks field could potentially reduce the total overall number of constraints.  An iteration of poseidon does get kicked off after a set number of elements are observed.

puma314 (approved):

PR #204, opened 2023-09-29
FunctionGatewayV2

puma314:


contracts/src/FunctionGatewayV2.sol
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        (bool status, ) = _callbackAddress.call(
+            abi.encodeWithSelector(_callbackSelector, _input, _output, _context)
+        );
+        if (!status) {
+            revert CallbackFailed(_callbackAddress, _callbackSelector);
+        }
+
+        delete requests[_requestId];
```
mattstam, line 60: `delete requests[_requestId];` needs to occur before the .call

contracts/src/FunctionGatewayV2.sol
```diff
@@ -0,0 +1,165 @@
+contract FunctionGatewayUpdated {
+    mapping(bytes32 => address) functionIdToVerifier;
+
+    function callbackRequestV1(
+        bytes32 _functionId,
```
jtguibas, line 5: I feel like a better name for this function is something like `requestWithContext`

jtguibas: and likewise `fulfillRequestWithContextV1`

contracts/src/FunctionGatewayV2.sol
```diff
@@ -0,0 +1,165 @@
...
+            abi.encodeWithSelector(_callbackSelector, _input, _output, _context)
+        );
+        if (!status) {
+            revert CallbackFailed(_callbackAddress, _callbackSelector);
+        }
+
+        delete requests[_requestId];
+    }
+
+    function storeRequestV1(
```
jtguibas, line 63: if most users are going to use this, we should just call it `requestV1` imo

contracts/src/FunctionGatewayV2.sol
```diff
@@ -0,0 +1,165 @@
...
+
+        if (_address != address(0)) {
+            _address.call(
+                abi.encodeWithSelector(_selector, _functionId, _input, _output)
+            );
+            delete oracle[_functionId][inputHash];
+        }
+    }
+
+    function verifyStored(
```
jtguibas, line 113: the name of this function should be on word, like `verify` or `check` or something

contracts/src/FunctionGatewayV2.sol
```diff
@@ -0,0 +1,165 @@
...
+                _proof
+            )
+        ) {
+            revert InvalidProof();
+        }
+
+        // Default case
+        oracle[_functionId][inputHash] = outputHash;
+
+        if (_address != address(0)) {
```
mattstam, line 105: IMO we should have the callback functionality in a separate function rather than shove it in here

contracts/src/FunctionGatewayV2.sol
mattstam, line 60: Not relevant to the changes but please remember to `delete requests[_requestId];` before the .call

contracts/src/FunctionGateway.sol
```diff
@@ -28,174 +29,200 @@ contract FunctionGateway is IFunctionGateway, FunctionRegistry, TimelockedUpgrad
...
 
-        // Update the request.
-        r.callbackFulfilled = true;
+        if (_callbackAddress == address(0)) {
+            // If there is no callback, then just store the result
+            verifiedResults[functionId][inputHash] = outputHash;
+            return;
+        }
+
+        VerifiedCall currentVerifiedCall = VerifiedCall({
```
mattstam, line 210: I'm really surprised if saving `currentVerifiedCall` to memory like this works, pretty sure it needs to be a storage slot.

PR #216, opened 2023-10-04
refactor: `CircuitVariable`

puma314:
## Overview
- Refactor `CircuitVariable` to remove previously slow `elements`, `from_elements` and `nb_elements` functions. 
- `CircuitVariable` implements `get`, `set` and `constant` in the trait. 
- Implementers of `CircuitVariable` will need to define `nb_elements`, `from_elements` and `elements`, or derive them from existing `CircuitVariable` implmentations.

### DIFF
--- a/plonky2x/src/frontend/eth/vars.rs
+++ b/plonky2x/src/frontend/eth/vars.rs
@@ -23,11 +23,16 @@ impl CircuitVariable for BLSPubkeyVariable {
         Self(BytesVariable::init_unsafe(builder))
     }
 
-    fn constant<L: PlonkParameters<D>, const D: usize>(
-        builder: &mut CircuitBuilder<L, D>,
-        value: Self::ValueType<L::Field>,
-    ) -> Self {
-        Self(BytesVariable::constant(builder, value))
+    fn nb_elements() -> usize {
+        BytesVariable::<48>::nb_elements()
+    }
+
+    fn elements<F: RichField>(value: Self::ValueType<F>) -> Vec<F> {
+        BytesVariable::<48>::elements(value)
+    }
+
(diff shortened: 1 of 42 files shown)


plonky2x/src/frontend/eth/vars.rs
```diff
@@ -23,11 +22,16 @@ impl CircuitVariable for BLSPubkeyVariable {
         Self(BytesVariable::init_unsafe(builder))
     }
 
-    fn constant<L: PlonkParameters<D>, const D: usize>(
-        builder: &mut CircuitBuilder<L, D>,
-        value: Self::ValueType<L::Field>,
-    ) -> Self {
-        Self(BytesVariable::constant(builder, value))
+    fn nb_elements() -> usize {
```
ratankaliani, line 25: nit: Is there a better way to do this?

plonky2x/src/frontend/eth/vars.rs
```diff
@@ -66,14 +62,16 @@ impl CircuitVariable for AddressVariable {
...
 
-    fn constant<L: PlonkParameters<D>, const D: usize>(
-        builder: &mut CircuitBuilder<L, D>,
-        value: Self::ValueType<L::Field>,
-    ) -> Self {
-        Self(BytesVariable::constant(
-            builder,
-            value.as_bytes().try_into().expect("wrong slice length"),
-        ))
+    fn nb_elements() -> usize {
```
ratankaliani, line 65: nit: Is there a better way to do this?

plonky2x/src/frontend/eth/vars.rs
puma314: I don't think so, this is what I also do

ratankaliani (approved):
lgtm!

PR #218, opened 2023-10-05
feat: mapreduce serializer generic + circuit loading speedups

jtguibas:


plonky2x/src/backend/function/mod.rs
```diff
@@ -174,39 +176,33 @@ impl<C: Circuit> VerifiableFunction<C> {
             AlgebraicHasher<InnerParameters::Field>,
         OuterParameters::Config: Serialize,
     {
-        let path = format!("{}/main.circuit", args.build_dir);
```
kevjue, line 177: nit:  Seems like there is alot of duplicated code between `prove` and `prove_wrapped`.  Might want to consider consolidating them a bit, if possible.

kevjue (commented):
Looks like CI formatting is failing because this import (https://github.com/succinctlabs/succinctx/blob/9352a433244318d850b8bcf2d1bc1b848b27846a/plonky2x/src/frontend/uint/uint32_n.rs#L334-L335) is not used here:  https://github.com/succinctlabs/succinctx/blob/9352a433244318d850b8bcf2d1bc1b848b27846a/plonky2x/src/frontend/uint/uint64.rs#L50.

kevjue (approved):
LGTM.

PR #230, opened 2023-10-10
feat: contracts v2

jtguibas:


contracts/src/FunctionGatewayV2.sol
```diff
@@ -0,0 +1,294 @@
...
+        emit Request(
+            nonce,
+            _functionId,
+            _input,
+            _context,
+            _callbackSelector,
+            _callbackGasLimit
+        );
+
+        // Increment the nonce.
```
mattstam, line 104: I think its better to increment before the saving & emitting the request. It's not a huge deal but (1) never have zero nonce request (2) if caller needs to check nonce again after zkRequest for some reason, it's for their own.

contracts/src/FunctionGatewayV2.sol
```diff
@@ -0,0 +1,294 @@
...
+
+        // Execute the callback.
+        (bool status, ) = _callbackAddress.call(
+            abi.encodeWithSelector(_callbackSelector, _output, _context)
+        );
+
+        // If the callback failed, revert.
+        require(status, "FunctionGateway: callback failed");
+
+        // Delete the callback hash for a gas refund.
```
mattstam, line 161: Delete before .call

contracts/src/FunctionGatewayV2.sol
```diff
@@ -0,0 +1,294 @@
...
+    ) external {
+        // Compute the input and output hashes.
+        bytes32 inputHash = sha256(_input);
+        bytes32 outputHash = sha256(_output);
+
+        // Verify the proof.
+        _verify(_functionId, inputHash, outputHash, _proof);
+
+        // Set the current verified call.
+        currentVerifiedFunctionId = _functionId;
```
mattstam, line 205: Maybe cleaner to just put this all in currentVerifiedCall struct

mattstam (commented):
Overall looks good!

contracts/src/mocks/MockFunctionGateway.sol
puma314, line 1: I don't think we should delete this file, we should update it to the latest thing

contracts/test/FunctionGateway.t.sol
puma314, line 1: Don't think we should delete this file either, but agree that writing tests is lower priority

contracts/src/interfaces/IFunctionGateway.sol
```diff
@@ -1,91 +1,56 @@
...
-        bytes32 outputsRoot,
-        bytes32 verificationKeyRoot
-    ) external;
-
-    function callback(bytes32 requestId, bytes memory output, bytes memory context) external;
-
-    function calculateFeeAmount() external view returns (uint256);
-
-    function calculateFeeAmount(uint256 gasLimit) external view returns (uint256);
+    function zkCall(
```
puma314, line 52: I think this should return `(bool ok, bytes memory)` to match the structure of `.call` normally, what do you think?

contracts/src/FunctionGateway.sol
```diff
@@ -2,246 +2,270 @@
...
-        uint256 feeAmount = _handlePayment(_gasLimit, _refundAccount, msg.sender, msg.value);
-
-        bytes32 requestId = keccak256(abi.encode(nonce, r));
-        requests[requestId] = r;
-
-        emit ProofRequested(nonce, _functionId, requestId, _input, _context, _gasLimit, feeAmount);
-        nonce++;
-        return requestId;
-    }
+        bytes32 contextHash = sha256(_context);
```
puma314, line 64: nit: here we should do `keccack` since it's cheaper (also in our circuit once we have an optimized keccack implementation, we should maybe do that for the input as well)?

contracts/src/FunctionGateway.sol
```diff
@@ -2,246 +2,270 @@
...
+        // Increment the nonce.
+        nonce++;
 
-        // Verify the proof.
-        address verifier = verifiers[r.functionId];
-        if (!IFunctionVerifier(verifier).verify(r.inputHash, _outputHash, _proof)) {
-            revert InvalidProof(address(verifier), r.inputHash, _outputHash, _proof);
-        }
+        // Send the fee to the vault.
+        IFeeVault(feeVault).depositNative{value: msg.value}(callbackAddress);
```
puma314, line 91: While we're at it, can we change the `FeeVault` so that it doesn't try to refund and fail if it isn't able to? It always borks out forge simulations and is super annoying

contracts/src/FunctionGateway.sol
```diff
@@ -2,246 +2,270 @@
...
-                revert RequestNotFound(requestId);
-            } else if (r.proofFulfilled) {
-                revert ProofAlreadyFulfilled(requestId);
-            }
-            inputHashes[i] = r.inputHash;
-            address verifier = verifiers[r.functionId];
-            verificationKeyHashes[i] = IFunctionVerifier(verifier).verificationKeyHash();
-        }
+        // Reconstruct the callback hash.
+        bytes32 contextHash = sha256(_context);
```
puma314, line 118: nit: change to keccack here

contracts/src/FunctionGateway.sol
```diff
@@ -2,246 +2,270 @@
...
-            revert ProofNotFulfilled(_requestId);
+        // If the callback failed, revert.
+        if (!status) {
+            revert CallbackFailed(_callbackSelector, _output, _context);
         }
 
-        // Update the request.
-        r.callbackFulfilled = true;
+        // Delete the callback hash for a gas refund.
+        delete requests[_nonce];
```
puma314, line 153: I think here you have to delete the requests[_nonce] before the external call for re-entrancy reasons?

contracts/src/FunctionGateway.sol
```diff
@@ -2,246 +2,270 @@
...
+        // Verify the proof.
+        _verify(_functionId, inputHash, outputHash, _proof);
 
-    /// @notice Calculates the feeAmount for the default gasLimit.
-    function calculateFeeAmount() external view returns (uint256 feeAmount) {
-        return calculateFeeAmount(DEFAULT_GAS_LIMIT);
-    }
+        // Set the current verified call.
+        verifiedFunctionId = _functionId;
+        verifiedInputHash = inputHash;
```
puma314, line 200: nit: weird spacing?

puma314 (commented):
Looks great, few small comments

contracts/src/FunctionGatewayV2.sol
jtguibas: resolved.

contracts/src/FunctionGateway.sol
```diff
@@ -2,246 +2,295 @@
...
-        bytes4 _callbackSelector,
         bytes memory _context,
-        uint256 _gasLimit,
-        address _refundAccount
-    ) public payable returns (bytes32) {
+        bytes4 _callbackSelector,
+        uint32 _callbackGasLimit
+    ) external payable returns (bytes32) {
+        // Compute the callback hash uniquely associated with this request.
         bytes32 inputHash = sha256(_input);
```
puma314, line 63: why are you hashing here? you should just shove it all into the _requestHash keccack, just shove all the bytes including `_input` and `context` in there

contracts/src/FunctionGateway.sol
```diff
@@ -2,246 +2,295 @@
...
-        // Verify the proof.
-        address verifier = verifiers[r.functionId];
-        if (!IFunctionVerifier(verifier).verify(r.inputHash, _outputHash, _proof)) {
-            revert InvalidProof(address(verifier), r.inputHash, _outputHash, _proof);
-        }
+        return requestHash;
+    }
 
-        emit ProofFulfilled(_requestId, _outputHash, _proof);
+    /// @dev Creates a proof request for a call. This function is equivalent to an off-chain
```
puma314, line 96: nit: "off-chain request through our API"

contracts/src/FunctionGateway.sol
```diff
@@ -2,246 +2,295 @@
...
-        // Do some sanity checks.
-        FunctionRequest storage r = requests[_requestId];
-        if (r.callbackAddress == address(0)) {
-            revert RequestNotFound(_requestId);
-        } else if (r.proofFulfilled) {
-            revert ProofAlreadyFulfilled(_requestId);
-        }
+        // Store the callback hash.
+        requests[nonce] = requestHash;
+        emit RequestCallback(
```
puma314, line 81: this should have the `callbackAddress`, no?

contracts/src/FunctionGateway.sol
puma314: actually i understand--it's so that `_input` and `context` don't conflict i guess since they're both unsized, nvrmind

contracts/src/FunctionGateway.sol
```diff
@@ -2,246 +2,295 @@
...
+    /// @dev Creates a proof request for a call. This function is equivalent to an off-chain 
+    ///      request.
+    /// @param _functionId The function identifier.
+    /// @param _input The function input.
+    /// @param _callbackAddress The address of the callback contract.
+    /// @param _callbackData The data for the callback function.
+    function requestCall(
+        bytes32 _functionId,
+        bytes memory _input,
+        address _callbackAddress,
```
puma314, line 105: nit: _address
nit: _calldata

contracts/src/FunctionGateway.sol
```diff
@@ -2,246 +2,295 @@
...
-                revert ProofAlreadyFulfilled(requestId);
-            }
-            inputHashes[i] = r.inputHash;
-            address verifier = verifiers[r.functionId];
-            verificationKeyHashes[i] = IFunctionVerifier(verifier).verificationKeyHash();
+    /// @dev If the call matches the currently verified function, returns the output. Otherwise,
+    ///      this function reverts.
+    /// @param _functionId The function identifier.
+    /// @param _input The function input.
+    function verifyCall(
```
puma314, line 124: `verifiedCall`

PR #234, opened 2023-10-12
fix: gnark plonky2 verifier

puma314:


plonky2x/core/src/frontend/vars/bytes32.rs
```diff
@@ -110,6 +110,20 @@ impl SSZVariable for Bytes32Variable {
     }
 }
 
+impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
+    // Returns a Bytes32Variable with the first `num_bits` set to 0.
+    pub fn zero_top_bits(&mut self, original: Bytes32Variable, num_bits: usize) -> Bytes32Variable {
```
tamirhemo, line 115: Just a remark but we should probably say somewhere that Bytes32Variable is big endian.

plonky2x/verifier/circuit.go
```diff
@@ -29,92 +27,70 @@ type Plonky2xVerifierCircuit struct {
...
 	OutputHash frontend.Variable `gnark:"outputHash,public"`
 
-	verifierChip *verifier.VerifierChip `gnark:"-"`
-	CircuitPath  string                 `gnark:"-"`
+	// Private inputs to the circuit
+	ProofWithPis variables.ProofWithPublicInputs
+	VerifierData variables.VerifierOnlyCircuitData
+
+	// Circuit configuration that is not part of the circuit itself.
+	CommonCircuitData types.CommonCircuitData `gnark:"-"`
```
tamirhemo, line 35: Is this a constant now?

tamirhemo (approved):

plonky2x/core/src/backend/function/mod.rs
```diff
@@ -27,8 +25,13 @@ const VERIFIER_CONTRACT: &str = include_str!("../../../../../assets/Verifier.sol
 /// `Plonky2xFunction`s have all necessary code for a circuit to be deployed end-to-end.
 pub trait Plonky2xFunction {
     /// Builds the circuit and saves it to disk.
-    fn build<L: PlonkParameters<D>, const D: usize>(args: BuildArgs)
-    where
+    fn build<
+        L: PlonkParameters<D>,
+        WrapperParameters: PlonkParameters<D, Field = L::Field>,
```
jtguibas, line 30: I like this change

plonky2x/core/src/backend/function/mod.rs
```diff
@@ -81,36 +89,32 @@ impl<C: Circuit> Plonky2xFunction for C {
             let contract_path = format!("{}/FunctionVerifier.sol", args.build_dir);
             let mut contract_file = File::create(&contract_path).unwrap();
 
-            let circuit_digest_bytes = circuit
+            info!("First building wrapper circuit to get the wrapper circuit digest...");
+            // The wrapper circuit digest will get saved in the Solidity smart contract,
+            // which will use this value as a public input `VerifierDigest` in the Gnark plonky2 verifier
+            let wrapped_circuit = WrappedCircuit::<L, WrapperParameters, D>::build(circuit);
+
+            let mut circuit_digest_bytes = wrapped_circuit
```
jtguibas, line 97: We should standardize the circuit_digest calculation because it's used in a lot of other places e.g., knowing the name to use when saving things to disk.

Maybe it can be a method on top of `CircuitBuild`.

plonky2x/core/src/backend/function/mod.rs
```diff
@@ -81,36 +89,32 @@ impl<C: Circuit> Plonky2xFunction for C {
             let contract_path = format!("{}/FunctionVerifier.sol", args.build_dir);
             let mut contract_file = File::create(&contract_path).unwrap();
 
-            let circuit_digest_bytes = circuit
+            info!("First building wrapper circuit to get the wrapper circuit digest...");
+            // The wrapper circuit digest will get saved in the Solidity smart contract,
+            // which will use this value as a public input `VerifierDigest` in the Gnark plonky2 verifier
```
jtguibas, line 94: nit: why is plonky2 lowercse but Gnark capitalized

plonky2x/core/src/backend/function/mod.rs
jtguibas: Including the .reverse() part

plonky2x/core/src/backend/function/mod.rs
```diff
@@ -81,36 +89,32 @@ impl<C: Circuit> Plonky2xFunction for C {
...
-                .collect::<Vec<u8>>();
-            let circuit_digest_hash = sha2::Sha256::digest(full_circuit_digest_bytes);
-            assert!(
-                circuit_digest_hash.len() <= 32,
-                "circuit digest must be <= 32 bytes"
-            );
-
+                .to_bytes();
+            // to_bytes() returns the representation as LE, but we want to save it on-chain
+            // as BE because that is the format of the public input to the Gnark plonky2 verifier.
```
jtguibas, line 104: nit: why is plonky2 lowercse but Gnark capitalized

plonky2x/core/src/backend/function/mod.rs
```diff
@@ -81,36 +89,32 @@ impl<C: Circuit> Plonky2xFunction for C {
...
+            circuit_digest_bytes.reverse();
+
+            // The VerifierDigest is stored on-chain as a bytes32, so we need to pad it with 0s
+            // to store it in the Solidity smart contract.
+            // Note that we don't need to do any sort of truncation of the top bits because the
+            // circuit digest already lives in the bn254 field because the WrappedCircuit config
```
jtguibas, line 110: nit: why is bn254 lowercase but next line it's capitalized

jtguibas: nit: if you want line breaks by sentence but with no extra space are unusual in this repo

either collapse to 100 cols or add another line between line breaks

plonky2x/core/src/frontend/vars/bytes32.rs
jtguibas: nit: instead of top, maybe you use the word `be`

jtguibas: Yes, alternatively, rename the function to `mask_be_bits`

plonky2x/core/src/backend/function/mod.rs
```diff
@@ -171,6 +175,8 @@ impl<C: Circuit> Plonky2xFunction for C {
         );
 
         if let PublicOutput::Bytes(output_bytes) = output {
+            // TODO: can optimize this by saving the wrapped circuit in build and loading it from a binary
```
jtguibas, line 178: Good observation--in practice, this may not impact performance at all as it doesn't take too long to build the wrapper circuits JIT.

plonky2x/core/src/backend/function/mod.rs
```diff
@@ -259,7 +265,6 @@ contract FunctionVerifier is IFunctionVerifier, Verifier {
             abi.decode(_proof, (uint256[2], uint256[2][2], uint256[2]));
 
         uint256[3] memory input = [uint256(CIRCUIT_DIGEST), uint256(_inputHash), uint256(_outputHash)];
-        input[0] = input[0] & ((1 << 253) - 1);
```
jtguibas, line 262: I agree with this change

plonky2x/core/src/backend/function/mod.rs
puma314: In this case, the circuit_digest calculation is actually a bit different because the element is a `Hash` which lives in a different field depending on the hasher...

jtguibas (approved):

plonky2x/core/src/backend/function/args.rs
```diff
@@ -5,6 +5,12 @@ use clap::{Parser, Subcommand};
 pub struct BuildArgs {
     #[arg(long, default_value = "./build")]
     pub build_dir: String,
+
+    // This argument is needed to generate the FunctionVerifier.sol contract that contains
+    // the verification smart contract for the gnark-plonky-2 verifier.
+    // This needs to match the argument in `ProveArgs`
```
jtguibas, line 11: nit: commenting format weird

PR #243, opened 2023-10-17
feat:  made BoolVariable, U32Variable, and U32Target safer

kevjue:


plonky2x/core/src/frontend/uint/uint32.rs
```diff
@@ -17,23 +17,34 @@ use crate::prelude::*;
 ///
 /// Under the hood, it is represented as a single field element.
 #[derive(Debug, Clone, Copy)]
-pub struct U32Variable(pub Variable);
+#[allow(clippy::manual_non_exhaustive)]
+pub struct U32Variable {
+    pub variable: Variable,
+    /// This private field is here to force all instantiations to go the methods below
+    _private: (),
```
jtguibas, line 24: clever

plonky2x/core/src/frontend/uint/uint32.rs
```diff
@@ -17,23 +17,34 @@ use crate::prelude::*;
 ///
 /// Under the hood, it is represented as a single field element.
 #[derive(Debug, Clone, Copy)]
-pub struct U32Variable(pub Variable);
+#[allow(clippy::manual_non_exhaustive)]
+pub struct U32Variable {
+    pub variable: Variable,
+    /// This private field is here to force all instantiations to go the methods below
```
jtguibas, line 23: nit: period

jtguibas (approved):

plonky2x/core/src/frontend/builder/boolean.rs
```diff
@@ -1,16 +1,17 @@
 use crate::backend::circuit::PlonkParameters;
 use crate::frontend::builder::CircuitBuilder;
 use crate::frontend::vars::{BoolVariable, Variable};
+use crate::prelude::CircuitVariable;
 
 impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
     pub fn _false(&mut self) -> BoolVariable {
         let zero = self.zero::<Variable>();
-        zero.into()
+        BoolVariable::from_variables_unsafe(&[zero])
```
puma314, line 9: Maybe we should have a convention that every time `unsafe` is used, there is a brief comment as to why that's okay? (even to make life easier for our auditors)

plonky2x/core/src/frontend/hash/poseidon/poseidon256.rs
```diff
@@ -115,7 +115,11 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
 
                 let hash_byte_vec = bit_list
                     .chunks(8)
-                    .map(|chunk| ByteVariable(array![i => BoolVariable::from(chunk[i].target); 8]))
+                    .map(|chunk| {
+                        ByteVariable::from_variables_unsafe(
```
puma314, line 119: Should we also make `ByteVariable` safer by adding the `_private()` field as well? I think a separate PR might be best for that as well

plonky2x/core/src/frontend/uint/uint32_n.rs
```diff
@@ -139,12 +139,12 @@ macro_rules! make_uint32_n {
                 let self_targets = self
                     .limbs
                     .iter()
-                    .map(|x| U32Target(x.0 .0))
+                    .map(|x| U32Target(x.variable.0))
```
puma314, line 142: Is `U32Target` safe? Or is the reasoning here that this method is okay because we make other assumptions about what is going into the `U32Target`?

plonky2x/core/src/frontend/uint/uint32_n.rs
```diff
@@ -139,12 +139,12 @@ macro_rules! make_uint32_n {
...
                     .collect::<Vec<_>>();
                 let rhs_targets = rhs
                     .limbs
                     .iter()
-                    .map(|x| U32Target(x.0 .0))
+                    .map(|x| U32Target(x.variable.0))
```
puma314, line 147: relatedly, i think we should make `U32Target` pub(crate) so that it's only accessible via U32Variables

plonky2x/core/src/frontend/builder/boolean.rs
kevjue: Ya, makes sense.

plonky2x/core/src/frontend/hash/poseidon/poseidon256.rs
kevjue: Actually, `from_variables_unsafe` is not needed here.  It's underlying elements are guaranteed to be booleans, so there is it's safe.  Will modify that.

Also, since ByteVariable's underlying variables are `BoolVariables`, it should be "safe" for someone to instantiate it directly.  The general principle that I did was to implement the `_private()` pattern for variables that could be be overflowed, and allow direct creation for all other ones.

This is actually somewhat similar to the principle used within plonky2.

plonky2x/core/src/frontend/uint/uint32_n.rs
kevjue: Good catch.  I didn't go through any Target based structs in the plonky2x.

plonky2x/core/src/frontend/vars/stream.rs
```diff
@@ -50,6 +50,9 @@ impl<L: PlonkParameters<D>, const D: usize> OutputVariableStream<L, D> {
     }
     pub fn read<V: CircuitVariable>(&self, builder: &mut CircuitBuilder<L, D>) -> V {
         let variables = self.read_exact(builder, V::nb_elements());
+
+        // Reads from stream don't do any validity checks on the circuit variable.  It is the
+        // stream reader's responsibility to do so.
         V::from_variables_unsafe(&variables)
```
puma314, line 56: Isn't this kind of bad? All places where we use streams, I think that we should do `V::from_variables` because when we're reading in `bytes` and stuff like that, isn't it good to have it be range-checked?

plonky2x/core/src/frontend/vars/stream.rs
```diff
@@ -97,6 +100,8 @@ impl VariableStream {
 
     pub fn read<V: CircuitVariable>(&mut self) -> V {
         let variables = self.0.read_exact(V::nb_elements());
+        // Reads from stream don't do any validity checks on the circuit variable.  It is the
+        // stream reader's responsibility to do so.
         V::from_variables_unsafe(variables)
```
puma314, line 105: same here, I think we should have this be `from_variables`

puma314 (approved):

PR #246, opened 2023-10-18
fix: read, read_unsafe, and add `_witness` for non-verified generator/hint result

kevjue:


puma314 (approved):

plonky2x/core/src/frontend/hash/keccak/mod.rs
```diff
@@ -11,7 +11,8 @@ pub mod keccak256;
 
 impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
     /// WARNING: DO NOT USE IN PRODUCTION, this is unconstrained!
-    pub fn keccak256(&mut self, bytes: &[ByteVariable]) -> Bytes32Variable {
+    pub fn keccak256_witness(&mut self, bytes: &[ByteVariable]) -> Bytes32Variable {
+        // TODO: Need to constrain generator result
```
ratankaliani, line 15: @puma314 Should we constrain this now?

plonky2x/core/src/frontend/eth/mpt/rlc.rs
```diff
@@ -35,7 +35,7 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
         b_offset: Variable,
         len: Variable,
     ) {
-        // TODO: implement
+        // TODO: Need to constrain generator result.
```
ratankaliani, line 38: @puma314 Should we constrain this now?

puma314: No that's fine, no one is going to use this file

plonky2x/core/src/frontend/hash/keccak/mod.rs
puma314: no, we don't have the implementation so fine to just leave as long as the name is changed to end with `_witness`

PR #251, opened 2023-10-23
feat: add `compress_point_variable`

ratankaliani:
- Mirror logic from `compress_point` to return a `Bytes32Variable` when compressing points.


kevjue (approved):
lgtm

PR #255, opened 2023-10-24
feat: add MockFunctionGateway and contract examples

mattstam:
Add `MockFunctionGateway`. An example of a `StorageOracle` Function that checks an account's storage slot value is also included, with a `NounsOwnership` contract that builds ontop of this.

This should make it clear how to use the FunctionGateway API, and test contracts utilizing the mock.

Test output:
```sh
Running 1 test for test/examples/NounsOwnership.t.sol:NounsOwnershipTest
[PASS] test_OwnerOf_WithMock() (gas: 211081)
Test result: ok. 1 passed; 0 failed; 0 skipped; finished in 673.75ms
```


contracts/src/mocks/MockFunctionGateway.sol
```diff
@@ -1,4 +1,184 @@
...
+    VmSafe private constant vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
+    uint32 public nonce;
+    bytes32 public verifiedFunctionId;
+    bytes32 public verifiedInputHash;
+    bytes public verifiedOutput;
+    bool public isCallback;
+    mapping(uint32 => bytes32) public requests;
+    mapping(bytes32 => bytes) public outputs;
+
+    function loadFixture(string memory _path) external {
```
puma314, line 23: This looks awesome! Can we also add another function that's `loadInputOutput(bytes memory input, bytes memory output)` so that people can load stuff inline as well?

contracts/src/mocks/MockFunctionGateway.sol
```diff
@@ -1,4 +1,184 @@
...
+            callbackAddress,
+            _callbackSelector,
+            _callbackGasLimit,
+            msg.value
+        );
+        nonce++;
+
+        // If fixture has been pre-loaded, automatically fulfill and callback.
+        bytes memory output = outputs[inputHash];
+        if (output.length > 0) {
```
puma314, line 73: I think sometimes a valid output can actually be 0 bytes--for example if your proof is just verifying that f(x) holds true if that makes sense.

contracts/src/mocks/MockFunctionGateway.sol
```diff
@@ -1,4 +1,184 @@
...
+                revert CallbackFailed(_callbackSelector, output, _context);
+            }
+
+            emit RequestFulfilled(nonce, _functionId, inputHash, outputHash);
+        }
+
+        return requestHash;
+    }
+
+    /// @dev This function only exists to implement interface. Use requestCall with your target
```
puma314, line 90: Should we have it revert if it only exists to implement the interface? I'm kind of confused haha why we have this in the interface anyways actually

contracts/src/examples/StorageOracle.sol
```diff
@@ -0,0 +1,101 @@
...
+            functionId, input, context, StorageOracle.handleStorageSlot.selector, DEFAULT_GAS_LIMIT
+        );
+
+        emit SlotRequested(blockNumber, blockHash, _account, _slot);
+    }
+
+    /// @dev Callback function to recieve the storage slot value from the FunctionGateway. If for existing slot, MUST update
+    ///      for a more recent blockNumber.
+    function handleStorageSlot(bytes memory _output, bytes memory _context) external {
+        if (msg.sender != gateway) {
```
puma314, line 86: nit: alos have to check `isCallback` on the gateway

contracts/src/examples/StorageOracle.sol
```diff
@@ -0,0 +1,101 @@
...
+            revert InvalidL1BlockHash();
+        }
+        uint256 blockNumber = L1_BLOCK.number();
+        if (blockNumber == 0) {
+            revert InvalidL1BlockNumber();
+        }
+
+        bytes memory input = abi.encode(blockHash, _account, _slot);
+        bytes memory context = abi.encode(blockNumber, _account, _slot);
+        requestId = IFunctionGateway(gateway).requestCallback{value: msg.value}(
```
puma314, line 76: might be nice to have a comment about what this `functionId` is doing

puma314 (commented):
Generally looks really great, thanks so much for doing this--it's extremely well documented and thoroughly tested and written very well! Some small comments

contracts/src/mocks/MockFunctionGateway.sol
mattstam: In the normal case, it seems fine for `Call` to have different requesting contract and receiving contract. We just have this hear as a convince to just make it one function by assuming the requestinad and receiving are the same.

contracts/src/examples/StorageOracle.sol
mattstam: fixed

contracts/src/examples/StorageOracle.sol
mattstam: added

contracts/src/mocks/MockFunctionGateway.sol
mattstam: Removed

contracts/src/mocks/MockFunctionGateway.sol
mattstam: Added

puma314 (approved):

PR #265, opened 2023-10-30
feat: latest changes for nouns

jtguibas:


plonky2x/core/src/frontend/uint/uint64.rs
```diff
@@ -50,6 +51,20 @@ make_uint32_n!(U64Variable, u64, 2);
 make_uint32_n_tests!(U64Variable, u64, 2);
 
 impl U64Variable {
+    /// Converts a U64Variable to Variable.
+    ///
+    /// Note: This function assumes that the U64 is in the range [0, 2^64-2^32+1). Otherwise, it
+    /// will overflow.
+    pub fn to_variable<L: PlonkParameters<D>, const D: usize>(
```
puma314, line 58: can we name this `to_variable_wrap` since that's what it is doing?

puma314 (approved):

plonky2x/core/src/frontend/uint/uint64.rs
jtguibas: called it to_variable_with_overflow

PR #267, opened 2023-10-30
feat: onlyProver for Gateway

mattstam:
`fulfillCallback` and `fulfillCall` can only be called by a prover.


contracts/src/upgrades/Versioned.sol
```diff
@@ -6,6 +6,6 @@ pragma solidity ^0.8.16;
 /// @dev This is used for easier management and version compatibility checking between contracts.
 abstract contract Versioned {
     function VERSION() external pure virtual returns (string memory) {
-        return "1.0.0";
+        return "1.0.1";
```
mattstam, line 9: Bump since we will do upgrade after merge.

jtguibas (approved):

PR #276, opened 2023-11-03
chore: clean up the RLP/MPT codebase and add some features

hidenori-shinohara:
This PR does two things primarily.

1. Refactor the RLP/MPT codebase for better readability and add more comments.
2. Enhance the features of the decode verifier. It now handles the case when the encoded input has 1 prefix byte, and the verifier now detects the incorrect prefix bytes.


plonky2x/core/src/frontend/eth/rlp/utils.rs
```diff
@@ -0,0 +1,379 @@
+use ethers::types::Bytes;
+use log::info;
+use num::bigint::ToBigInt;
+use num::BigInt;
+
+/// Byte array of at most 32 bytes, potentially padded. This could be a hash, value,
+/// rlp.encode(branch node), rlp.encode(leaf node), or rlp.encode(extension node) under the hood.
+pub const MAX_STRING_SIZE: usize = 32;
+pub type FixedSizeString = [u8; MAX_STRING_SIZE];
```
jtguibas, line 9: Perhaps FixedSizeString should actually be the following and encode both the data and the length.

struct FixedSizeString<const L: usize> {
    data: [u8; L];
    len: usize
};

jtguibas: And then maybe you should make an impl on the struct such that you can do

let s: FixedSizeString;
let metadata: RLPEncodingMetadata = s.rlp_encoding_metadata();

jtguibas: In general, we should try to make things more object oriented

plonky2x/core/src/frontend/eth/rlp/utils.rs
```diff
@@ -0,0 +1,379 @@
...
+///    which is exactly 32 bytes.
+pub const MAX_NODE_SIZE: usize = 17;
+pub type FixedSizeMPTNode = [FixedSizeString; MAX_NODE_SIZE];
+
+/// This represents the length of each string in a node, potentially padded Since each
+/// FixedSizeString in FixedSizeString is padded, we need to keep track of the "true" length of each
+/// string.
+pub type FixedSizeStringLengths = [usize; MAX_NODE_SIZE];
+
+pub fn bool_to_u32(b: bool) -> u32 {
```
jtguibas, line 30: this function is unecssary, i think you can do `b as u32`

plonky2x/core/src/frontend/eth/rlp/utils.rs
```diff
@@ -0,0 +1,379 @@
...
+    if b {
+        1
+    } else {
+        0
+    }
+}
```
jtguibas, line 36: https://github.com/succinctlabs/optimism-bedrock-contracts/blob/main/rlp/RLPReader.sol

We should follow the conventions in this file and have RLPItem etc. for even the normal functions?

plonky2x/core/src/frontend/eth/rlp/decoder.rs
```diff
@@ -0,0 +1,249 @@
+//! This file implements RLP decoder.
+//!
+//! Reference: https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/
+//!
+//! We have a total of 6 different ways of representing Merkle Patricia Tries.
+//! - Encoded versions
+//!     1. Encoded MPT : vec<u8>
+//!     2. Padded Encoded MPT : ([u8; usize], usize)
+//!     3. Circuita Padded Encoded MPT : ArrayVariable<ByteVariable, ENCODING_LEN>
```
jtguibas, line 9: Should this be a 2D array like 2?

plonky2x/core/src/frontend/eth/rlp/decoder.rs
```diff
@@ -0,0 +1,249 @@
...
+//! - Decoded versions
+//!     1'. MPT : RLPItem
+//!     2'. Padded MPT : MPTNodeFixedSize
+//!     3'. Circuit Padded MPT : ArrayVariable<ArrayVariable<ByteVariable, 32>, 17>
```
jtguibas, line 13: We can make a new type here called like MPTFixedSizeVariable

plonky2x/core/src/frontend/eth/rlp/decoder.rs
```diff
@@ -0,0 +1,249 @@
...
+//! builder.rs.
+
+/// An item is a string (i.e., byte array) or a list of items.
+#[derive(PartialEq, Debug)]
+pub enum RLPItem {
+    String(Vec<u8>),
+    List(Vec<RLPItem>),
+}
+
+/// Private helper struct to iterate over a vector.
```
jtguibas, line 38: Why not just use Stream<u8>?

plonky2x/core/src/frontend/eth/rlp/utils.rs
```diff
@@ -0,0 +1,293 @@
...
+//! Reference 1: https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/
+//! Reference 2: https://ethereum.org/en/developers/docs/data-structures-and-encoding/patricia-merkle-trie
+
+use num::bigint::ToBigInt;
+use num::BigInt;
+
+use super::decoder::RLPItem;
+use crate::frontend::eth::rlp::decoder::decode;
+
+pub const MAX_RLP_ITEM_SIZE: usize = 32;
```
jtguibas, line 12: space here

plonky2x/core/src/frontend/eth/rlp/decoder.rs
hidenori-shinohara: I couldn't figure out a way to track the number of bytes we have read from this stream. The information is necessary for the recursive case. When recursively calling this function in a list, RLP only tells us how many _bytes_ we'd need to read. But let me know if there's a way to do so. I don't like having a new private iterator class just for this purpose.

plonky2x/core/src/frontend/eth/rlp/decoder.rs
hidenori-shinohara: Oh okay! I like that. Maybe in the next PR?

plonky2x/core/src/frontend/eth/rlp/decoder.rs
hidenori-shinohara: I added "Variable" that denotes the length, which I assume is what you meant by 2D array? (I did the same for `3'` as well.)

plonky2x/core/src/frontend/eth/rlp/decoder.rs
hidenori-shinohara: oh I completely misunderstood! `Stream` is in our codebase. For whatever reason, I thought it was a native Rust class 🤦  Yes, that makes perfect sense. I'll do that.

hidenori-shinohara: Fixed!

hidenori-shinohara:
image

For future reference...

plonky2x/core/src/frontend/eth/rlp/utils.rs
```diff
@@ -0,0 +1,288 @@
...
+        return MPTNodeFixedSize::default();
+    }
+    MPTNodeFixedSize::from(decode(&encoded[..len]))
+}
+
+/// This calculates the prefix and the length of the encoding that we would get if we were to encode
+/// the given string.
+///
+/// More specifically, the first return value is the prefix of `rlp_encode(padded_string[..len])`.
+/// The econd return value is `rlp_encode(padded_string[..len]).len()`. This function is
```
puma314, line 108: nit: second

puma314 (approved):

plonky2x/core/src/frontend/eth/mpt/builder.rs
```diff
@@ -77,7 +77,7 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
         root: Bytes32Variable,
         value: Bytes32Variable,
     ) {
-        const ELEMENT_LEN: usize = 34; // Maximum size of list element
+        const ELEMENT_LEN: usize = 32; // Maximum size of list element
```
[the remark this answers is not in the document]
hidenori-shinohara: This is something that I'm not 100% sure about. The tests do indeed pass, but I'm not convinced that my change works perfectly fine with the real world case. I figured that this change is okay since we'll inevitably revisit this while I'll work on the next few PRs. This PR ignores other cases that may or may not occur in the real world, anyway. (e.g., I'm not 100% sure if a branch node with a node inlined actually exists)

For future reference, partial conclusions from my reading are:

- The vast majority of the time it will indeed be <= 32 bytes.
    - A branch node generally should have either a null or rep-encoding of a node smaller than 32 bytes or a 32-byte hash.
    - A path in a leaf node or an extension node is capped at 32 bytes as the max path length is 64 characters = 32 bytes. I'm fairly convinced that this is going to be much smaller than 32 bytes given the size of Ethereum in general.
    - The `key` field in an extension node is always going to be 32 bytes as they use keccak256. 
- The only one that isn't documented in the RLP/MPT specs is the `value` field in a leaf node. @jtguibas and I discussed that it's probably capped at 32 bytes.

plonky2x/core/src/frontend/eth/rlp/builder.rs
```diff
@@ -1,213 +1,56 @@
...
         let len = input_stream.read_value::<Variable>();
         let finish = input_stream.read_value::<BoolVariable>();
 
-        let (decoded_list, decoded_list_lens, len_decoded_list) =
-            decode_element_as_list::<ENCODING_LEN, LIST_LEN, ELEMENT_LEN>(
-                &encoded,
-                len.as_canonical_u64() as usize,
-                finish,
-            );
+        let decoded = decode_padded_mpt_node::<ENCODING_LEN, LIST_LEN>(
```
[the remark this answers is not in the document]
hidenori-shinohara: After some thoughts, I decided that we can just remove it as this function's actually really not using that information. This function doesn't need to know the maximum size and `len` gives the true length.


The check regarding whether the given list is length 2 or 17 can be done inside `MPTNodeFixedSize::from` as that should be responsible for taking a valid `RLPItem` and forming a valid `MPTNodeFixedSize`. And I think that's a better  abstraction.

plonky2x/core/src/frontend/eth/rlp/decoder.rs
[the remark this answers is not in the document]
hidenori-shinohara: yay!

plonky2x/core/src/frontend/eth/rlp/utils.rs
```diff
@@ -0,0 +1,288 @@
...
+/// This decodes an padded encoding of MPT node and returns the padded, decoded MPT node.
+///
+/// The input is a tuple of:
+/// - `encoded`: padded, RLP-encoded MPT node,
+/// - `len`: "true length" of `encoded`,
+/// - `finish`: a boolean indicating whether we should terminate early.
+pub fn decode_padded_mpt_node<const ENCODING_LEN: usize, const LIST_LEN: usize>(
+    encoded: &[u8],
+    len: usize,
+    finish: bool,
```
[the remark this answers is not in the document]
hidenori-shinohara: I renamed it to `skip_computation`.

plonky2x/core/src/frontend/eth/rlp/utils.rs
hidenori-shinohara: good catch!

PR #279, opened 2023-11-07
feat: `succinctx` offchain requests

ratankaliani:
Add support for off-chain requests to SuccinctX to a new `succinct-client` crate.


api/src/request.rs
```diff
@@ -0,0 +1,41 @@
+use ethers::types::Bytes;
+use log::{error, info};
+use serde::{Deserialize, Serialize};
+
+#[allow(non_snake_case)]
+#[derive(Serialize, Deserialize)]
+/// Struct to hold the data to be sent to the platform API.
```
jtguibas, line 7: struct SuccinctProofSerivce {
   rpc_url;
}


impl SuccinctProofService {
   fn new()
}
fn request_call(chainid, to, data, functionid, input)

api/src/request.rs
```diff
@@ -0,0 +1,60 @@
...
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+
+#[allow(non_snake_case)]
+#[derive(Serialize, Deserialize)]
+/// Data to be sent to the platform API with an offchain request.
+pub struct OffchainInput {
+    /// The chain id of the network to be used.
+    chainId: u32,
+    /// The address of the contract to call.
```
jtguibas, line 12: ethers::types::address

api/src/request.rs
```diff
@@ -0,0 +1,60 @@
...
+    to: Bytes,
+    /// The calldata to be used in the contract call.
+    data: Bytes,
+    /// The Succinct X function id to be called.
+    functionId: Bytes,
```
jtguibas, line 17: H256

api/src/request.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// The input to be used in the Succinct X function call.
+    input: Bytes,
+}
+
+/// Client to interact with the platform API.
+pub struct PlatformClient {
```
jtguibas, line 23: maybe call it SuccinctClient?

api/src/request.rs
ratankaliani: `H256` and `Address` serialize in a weird way

Ex. for `H256` the `to_string` is 0xaa...bb, instead of 0xaaaabb

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["plonky2x/core", "plonky2x/derive", "rustx"]
+members = ["plonky2x/core", "plonky2x/derive", "rustx", "api"]
```
jtguibas, line 2: This is not really an API, it's a client to communicate to an api

ratankaliani: `succinctx-client`?

jtguibas:
I see. In that case I still think it’s better if the function takes in the
arguments with the proper types and then inside the function body you cast
it.


Also, it’s almost universally better to avoid passing in a single large
struct into a function (instead flatten it to multiple arguments), unless
you expect that struct to be used by a user somewhere else.

Cargo.toml
ratankaliani: Updated to `succinctx-client`

api/src/request.rs
puma314: Can't you customize the serialization yourself to solve this?

api/src/request.rs
```diff
@@ -0,0 +1,75 @@
...
+        let res = self
+            .client
+            .post(request_url)
+            .header("Content-Type", "application/json")
+            .body(serialized_data)
+            .send()
+            .await
+            .unwrap();
+
+        // Check if the request was successful.
```
puma314, line 68: Here we should return a `Result` and I believe our endpoint returns a `request_id`, so we should return that in the `ok` version

puma314 (approved):
`client` sgtm

api/src/request.rs
ratankaliani: Solved serialization, had to add `serde` as a feature on the `alloy-primitives` crate.

PR #280, opened 2023-11-07
refactor: FunctionRegistry to use salt instead of name

mattstam:
```solidity
function deployAndRegisterFunction(address _owner, bytes memory _bytecode, string memory _name)
```
to

```solidity
function deployAndRegisterFunction(address _owner, bytes memory _bytecode, bytes32 _salt)
```


contracts/test/verifiers/VerifierPlonkRangeCheck.sol
```diff
@@ -16,7 +16,7 @@
 
 // Code generated by gnark DO NOT EDIT
 
-pragma solidity ^0.8.19;
+pragma solidity ^0.8.16;
```
mattstam, line 19: unrelated: need everything to be 0.8.16 to build with this compiler version

puma314 (approved):

PR #286, opened 2023-11-10
fix: eddsa + sha512 pad fixes

ratankaliani:
- `curta_eddsa_verify_sigs_conditional`

- Add test for `sha512_pad`. Previously, the `max_num_chunks` was calculated incorrectly.


tamirhemo (approved):
lgtm

PR #289, opened 2023-11-18
chore: change the visibility of num mod

ChengYueJia:


puma314:
@ChengYueJia is there a reason that you want `num` to be public? We intentionally made it not public so that all methods have to go through our `uint` functions, since the `num` functions are generally a bit ugly and shouldn't be used directly.

ChengYueJia:
> @ChengYueJia is there a reason that you want `num` to be public? We intentionally made it not public so that all methods have to go through our `uint` functions, since the `num` functions are generally a bit ugly and shouldn't be used directly.

Hi, as we need to use `plonky2x::frontend::uint::num::u32::gadgets::multiple_comparison::list_lte_circuit`   .

puma314:
@ChengYueJia if you continue to need this, we can potentially discuss merging it (or potentially making just `list_lte_circuit` gadget pub mod) since we really don'd want to make the entire `num` folder pub mod since it is not really meant to be used.

PR #295, opened 2023-12-01
feat(plonky2x): use `curta_sha256` in wrapper

ratankaliani:
## Overview
- Use `curta_sha256` in `plonky2x` wrapper to handle large public inputs better.
   - On Mac M2 Max for wrapping now takes ~4s, vs. ~2s before.
- Add `prove_with_partial_witness` and `prove_with_partial_witness_async` to `Plonky2x` circuits, allowing plonky2 `PartialWitness` to be used to generate proofs directly, without using `plonky2x` IO.

### DIFF
--- a/plonky2x/core/src/backend/circuit/build.rs
+++ b/plonky2x/core/src/backend/circuit/build.rs
@@ -40,9 +40,11 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuild<L, D> {
         PublicInput::new(&self.io)
     }
 
+    /// Generates a proof for the circuit using a plonky2 partial witness. The proof can be verified
+    /// using `verify`.
     pub fn prove_with_partial_witness(
         &self,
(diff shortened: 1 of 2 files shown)


plonky2x/core/src/backend/circuit/build.rs
```diff
@@ -40,10 +40,9 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuild<L, D> {
         PublicInput::new(&self.io)
     }
 
-    /// Generates a proof for the circuit. The proof can be verified using `verify`.
-    pub fn prove(
+    pub fn prove_with_partial_witness(
         &self,
-        input: &PublicInput<L, D>,
+        inputs: PartialWitness<L::Field>,
```
puma314, line 45: can you call this 'partial_witness', I think the variable naming is a bit incorrect

puma314 (approved):

PR #299, opened 2023-12-12
refactor: consolidate curta hash traits

kevjue:


plonky2x/core/src/backend/circuit/serialization/hints.rs
```diff
@@ -373,29 +372,35 @@ where
...
         r.register_hint::<EcOpProofHint>();
         r.register_async_hint::<Async<EcOpProofHint>>();
 
         r.register_hint::<EcOpResultHint>();
         r.register_async_hint::<Async<EcOpResultHint>>();
 
-        r.register_hint::<BLAKE2BDigestHint>();
-        r.register_async_hint::<Async<BLAKE2BDigestHint>>();
+        // r.register_hint::<BLAKE2BDigestHint>();
+        // r.register_async_hint::<Async<BLAKE2BDigestHint>>();
```
jtguibas, line 400: should we just delete these?

plonky2x/core/Cargo.toml
```diff
@@ -14,7 +14,7 @@ ci = []
 
 [dependencies]
 plonky2 = { git = "https://github.com/mir-protocol/plonky2.git", default-features = false }
-curta = { git = "https://github.com/succinctlabs/curta.git" }
+curta = { git = "https://github.com/succinctlabs/curta.git",  branch="kevjue/blake2b_int_types" }
```
jtguibas, line 17: can you change this to main before merging?

jtguibas:
looks good

plonky2x/core/src/frontend/hash/curta/mod.rs
```diff
@@ -131,14 +156,20 @@ pub trait SHA<L: PlonkParameters<D>, const D: usize, const CYCLE_LEN: usize>:
...
-                let mut flag = builder.constant::<BoolVariable>(true);
+                let mut t_var = builder.constant::<U32Variable>(0);
+                let chunk_size = builder.constant::<U32Variable>(128);
                 for j in 0..total_number_of_chunks {
+                    if HAS_T_VALUES {
+                        t_var = builder.add(t_var, chunk_size);
+                    }
                     let j_var = builder.constant::<U32Variable>(j as u32);
-                    let lte = builder.lte(last_chunk_index, j_var);
-                    let lte_times_flag = builder.and(lte, flag);
```
jtguibas, line 138: can you explain what's going on here? why can you just remove these lines?

jtguibas (commented):
are the t values constrained? i remember we forgot that last time

plonky2x/core/src/frontend/hash/curta/mod.rs
kevjue: I refactored the original logic (which I think is now easier to read).  So just to give some context, this loop is calculating the digest_bits (an array of bits to label which message chunk is the last chunk of the message) and t_values arrays.  Those arrays will be part of the curta public input.

The original logic is similar to:

```
digest_bits = []
flag = true
for j in range(0, num_chunks):
    if j <= last_chunk_idx AND flag:
        digest_bits[j] = false
    if j > last_chunk_idx:
        flag = false
```

I changed it to:
```
digest_bits = []
t_values = []
t_val = 0
for j in range(0, num_chunks):
    at_digest_bit = (j == last_chunk_idx)
    digest_bits[j] = at_digest_bit
 
   t_val += chunk_size
   if at_digest_bit:
      t_val = msg_length
   
   t_values[j] = t_val
```

kevjue:
> are the t values constrained? i remember we forgot that last time

Ya, the t_values are calculated within the circuit.

jtguibas (approved):

PR #309, opened 2024-01-06
feat: Local version of client

puma314:


client/src/request.rs
```diff
@@ -37,13 +47,153 @@ pub struct SuccinctClient {
...
+                "prove",
+                &input_file,
+                "--build-dir",
+                &build_dir,
+                "--wrapper-path",
+                &wrapper_binary,
+            ])
+            .env("PROVER", "local")
+            .env("RUST_LOG", "info")
+            .stdout(Stdio::piped())
```
ctian1, line 124: you can do Stdio::inherit to just have stdout/stderr print out

ctian1 (approved):

PR #310, opened 2024-01-06
feat: custom prover whitelist 

puma314:
Function owners can now specify if they want to use the default prover, a custom prover that they specify, or disable whitelisted proving altogether.

Audit Issue: https://veridise.notion.site/Contracts-Centralization-Risk-ec97d40516f3487eb907f2bd01f6135c Different from the recommended solution, but this solution was also with keeping in mind with some feedback from users.

### DIFF
--- a/contracts/src/SuccinctGateway.sol
+++ b/contracts/src/SuccinctGateway.sol
@@ -283,6 +283,17 @@ contract SuccinctGateway is ISuccinctGateway, FunctionRegistry, TimelockedUpgrad
         emit Call(_functionId, inputHash, outputHash);
     }
 
+    /// @notice Sets the whitelist status for a function.
+    /// @param _functionId The function identifier.
+    /// @param _status The whitelist status to set.
+    function setWhitelistStatus(bytes32 _functionId, WhitelistStatus _status) external {
+        if (msg.sender != verifierOwners[_functionId]) {
+            revert NotFunctionOwner(msg.sender, verifierOwners[_functionId]);
+        }
+        whitelistStatus[_functionId] = _status;
+        emit WhitelistStatusUpdated(_functionId, _status);
+    }
+
     /// @notice Add a custom prover.
     /// @param _functionId The function identifier.
     /// @param _prover The address of the prover to add.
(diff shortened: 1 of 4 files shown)


contracts/src/SuccinctGateway.sol
```diff
@@ -29,8 +35,10 @@ contract SuccinctGateway is ISuccinctGateway, FunctionRegistry, TimelockedUpgrad
     /// @notice A flag that indicates whether the contract is currently making a callback.
     bool public isCallback;
 
+    mapping(bytes32 => WhitelistStatus) public whitelistStatus;
```
mattstam, line 38: Need a way for function owners to update this, otherwise always stuck on default.

contracts/src/SuccinctGateway.sol
```diff
@@ -282,16 +298,32 @@ contract SuccinctGateway is ISuccinctGateway, FunctionRegistry, TimelockedUpgrad
 
     /// @notice Add the specified prover.
     /// @param _prover The address of the prover to add.
+    function addCustomProver(bytes32 functionId, address _prover) external {
```
mattstam, line 301: It would be better to put this function in FunctionRegistry. Would need to move the mapping into there but thats fine.

contracts/src/SuccinctGateway.sol
```diff
@@ -282,16 +298,32 @@ contract SuccinctGateway is ISuccinctGateway, FunctionRegistry, TimelockedUpgrad
...
     function addProver(address _prover) external onlyGuardian {
-        allowedProvers[_prover] = true;
-        emit ProverUpdated(_prover, true);
+        allowedProvers[bytes32(0)][_prover] = true;
+        emit ProverUpdated(bytes32(0), _prover, true);
     }
 
     /// @notice Remove the specified prover.
     /// @param _prover The address of the prover to remove.
+    function removeCustomProver(bytes32 functionId, address _prover) external {
```
mattstam, line 316: Same with this function.

contracts/src/SuccinctGateway.sol
```diff
@@ -282,16 +298,32 @@ contract SuccinctGateway is ISuccinctGateway, FunctionRegistry, TimelockedUpgrad
...
+        if (msg.sender != verifierOwners[functionId]) {
+            revert NotFunctionOwner(msg.sender, verifierOwners[functionId]);
+        }
+        allowedProvers[functionId][_prover] = false;
+        emit ProverUpdated(functionId, _prover, false);
+    }
+
     function removeProver(address _prover) external onlyGuardian {
```
mattstam, line 324: These functions could be renamed to addDefaultProver() and removeDefaultProver()

jtguibas (approved):

PR #328, opened 2024-01-12
feat(succinct-client): local relaying

ratankaliani:
Add local relaying of proofs to the `SuccinctClient`. For now, local relaying is only supported if local proving is enabled.

### DIFF
--- a/client/src/request.rs
+++ b/client/src/request.rs
@@ -396,8 +396,8 @@ impl SuccinctClient {
             let relayer_address = client.address();
             info!("Relayer address: {}", relayer_address);
 
-            let nonce = contract.nonce().call().await?;
-            info!("Nonce: {}", nonce);
+            let version = contract.version().call().await?;
+            info!("Version: {}", version);
 
             // Submit the proof to the Succinct X API.
             let tx = contract
--- a/client/src/request.rs
+++ b/client/src/request.rs
@@ -370,6 +370,7 @@ impl SuccinctClient {
             }
 
             let succinct_proof_data: SuccinctProofData = serde_json::from_value(proof_json)?;
+            let wallet = wallet.with_chain_id(succinct_proof_data.chain_id);
 
             let provider =
                 Provider::<Http>::try_from(ethereum_rpc_url).expect("could not connect to client");
--- a/client/src/request.rs
+++ b/client/src/request.rs
@@ -377,10 +377,15 @@ impl SuccinctClient {
 
             let address = get_gateway_address(succinct_proof_data.chain_id);
             // If gateway_address is provided, use that instead of the canonical gateway address.
-            let gateway_address = gateway_address.or(address).expect(
+            let mut gateway_address = gateway_address.or(address).expect(
                 "Gateway address must be provided when relaying a proof in local mode
                 if the chain does not have a canonical gateway address.",
             );
+
+            // Strip the 0x prefix from the gateway address.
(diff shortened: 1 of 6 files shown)


client/src/request.rs
```diff
@@ -267,4 +286,78 @@ impl SuccinctClient {
...
+                function_id,
+                input.clone(),
+            );
+        }
+
+        self.submit_platform_request(chain_id, to, calldata, function_id, input)
+            .await
+    }
+
+    pub async fn submit_proof(&self, ethereum_rpc_url: String, request_id: String) -> Result<()> {
```
puma314, line 312: Should we have another flag in the client that is like `RELAY=true` or something like that where if you're on local mode and relay=true, then after the proof generates, it automatically calls `submit_proof` or something like this? 

Btw I think that the `ethereum_rpc_url` should be read from the .env based on what chain_id is provided in the original request and panic if it is not found. Also for the gateway address, we should provide an option for people to specify that in their .env and fallback to our defaults if it exists since there might be chains where people deploy their own gateways.

ratankaliani: Thought about this, I don't think chain id should be read from `.env`, people might have their `.env` formatted differently, and forcing people to have things like `RPC_URL` in their `.env` in our format doesn't seem correct.

I'll add the `RELAY_MODE`.

client/src/utils.rs
```diff
@@ -0,0 +1,16 @@
+// List of Succinct Gateway addresses for chains.
+const GATEWAY_ADDRESSES: [(u32, &str); 6] = [
```
puma314, line 2: lol why not make this a `HashMap`?

client/src/request.rs
```diff
@@ -267,4 +306,115 @@ impl SuccinctClient {
...
+                    "Proof file {} does not exist.",
+                    proof_file
+                )));
+            }
+
+            // If it exists, attempt to submit the proof.
+            let proof_data = fs::read_to_string(proof_file)?;
+            let mut proof_json: serde_json::Value = serde_json::from_str(&proof_data)?;
+
+            // Strip alphanumeric characters from each of the proof_data fields (some are formatted
```
puma314, line 365: Is this something that we could do when we save the proof so that we don't have to do this sort of hack?

client/src/request.rs
```diff
@@ -267,4 +306,115 @@ impl SuccinctClient {
...
+    // TODO: Add support for hosted proving + local relaying.
+    pub async fn relay_proof(
+        &self,
+        request_id: String,
+        ethereum_rpc_url: Option<&str>,
+        wallet: Option<LocalWallet>,
+        gateway_address: Option<&str>,
+    ) -> Result<()> {
+        // If local mode, submit proof from local directory at proofs/output_{request_id}.json
+        if self.local_relay_mode {
```
puma314, line 346: Why would you ever call this function if `self.local_relay_mode` is false, I don't see the point of just wrapping the entire function in the flag?

client/src/utils.rs
ratankaliani: Not static, thought it was cleaner this way

client/src/request.rs
ratankaliani: Yeah this should be local prove mode

client/src/request.rs
ratankaliani: Yes

puma314 (approved):

PR #351, opened 2024-02-01
fix(audit): `get_fixed_subarray`

ratankaliani:


plonky2x/core/src/frontend/vars/array.rs
```diff
@@ -180,7 +180,44 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+        let array_variables = ArrayVariable::<Variable, ARRAY_SIZE>::from(
+            array_bytes
+                .as_slice()
+                .iter()
+                .map(|x| x.to_variable(self))
+                .collect_vec(),
+        );
+
+        let mut seed = Vec::new();
+        seed.extend_from_slice(array_bytes.as_slice());
```
puma314, line 200: In this case, shouldn't the seed also use the `array_variables`? and potentially also the `subarray_bytes` converted to variables? otherwise there will be 8x the number of things being poseidon-ed

ratankaliani: In `get_fixed_subarray_unsafe`, it packs 7 `ByteVariable`'s into a single `Variable` (56 bits), as it knows what the size of the targets are. `array_variables` == `array_bytes`, just as `Variable`'s instead of `ByteVariable`'s.

If we were to pass in `Variable`'s as the seed to `get_fixed_subarray_unsafe`, there wouldn't be a simple way to pack the `seed`. 

The logic in `get_fixed_subarray_unsafe` that compresses the chunk of `ByteVariable`'s into a single `Variable` before poseidon hashing is here: https://github.com/succinctlabs/succinctx/blob/254cffcdd48d4e831eb70332032d4e324aff1d3e/plonky2x/core/src/frontend/vars/array.rs#L255-L265

puma314: Oh I see, I didn't realize the packing was actually already going on in the unsafe function! that makes sense to me

puma314 (approved):

PR #357, opened 2024-02-12
chore: update dependency from curta to starkyx

mattstam:


ratankaliani (approved):
lgtm, as long as it builds/compiles correctly

PR #358, opened 2024-02-12
fix: handle chunk aligned messages for blake2b

kevjue:
This PR address two issues with blake2b related hash functions:

1) Fixed the case when the message is chunk aligned and not empty.  Basically, when the circuit was calculating the number of chunks of a message (or relatedly the idx of the last chunk), it was doing msg_len / chunk_len.  But if the msg_len is non zero and a multiple of chunk_len, that quotient value should be decremented by one.
2) In the pad_circuit_variable_length function (which is used when handling a blake2b request with variable size, it needed to replace all bytes after the inputted message len with zeros

### DIFF
--- a/plonky2x/core/src/frontend/hash/blake2/curta.rs
+++ b/plonky2x/core/src/frontend/hash/blake2/curta.rs
@@ -248,27 +248,21 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
         &mut self,
         input_byte_length: U32Variable,
     ) -> U32Variable {
-        let chunk_size = self.constant::<U32Variable>(128);
-        let last_chunk_idx = self.div(input_byte_length, chunk_size);
-
-        // Check to see if the input length is a multiple of 128 and it's not zero length. If it is,
-        // we need to subtract 1 from the last chunk index.
-
-        // Find out if the input length is a multiple of 128.
-        let mut tmp = self.mul(chunk_size, last_chunk_idx);
-        tmp = self.sub(tmp, input_byte_length);
-        let is_multiple = self.is_zero(tmp.variable);
-
-        // Find out if the input length is not zero.
-        let is_zero_length = self.is_zero(input_byte_length.variable);
-        let is_not_zero_length = self.not(is_zero_length);
-
-        // Find out if we need to subtract 1 from the last chunk index.
-        let do_minus_one = self.and(is_multiple, is_not_zero_length);
+        let zero = self.zero();
         let one = self.one();
-        let last_chunk_idx_minus_one = self.sub(last_chunk_idx, one);
 
-        self.select(do_minus_one, last_chunk_idx_minus_one, last_chunk_idx)
+        // The formula for calculating the last Blake2b chunk index is:
+        //  if input_byte_length == 0 { last_chunk_idx = 0 }
+        //  else { last_chunk_idx = (input_byte_length - 1) / 128 }
+        let is_zero_length = self.is_zero(input_byte_length.variable);
+
+        let chunk_size = self.constant::<U32Variable>(128);
+        // input_byte_length - 1
+        let input_byte_length_minus_one = self.sub(input_byte_length, one);
+        // (input_byte_length - 1) / 128
+        let last_chunk_idx = self.div(input_byte_length_minus_one, chunk_size);
+
+        self.select(is_zero_length, zero, last_chunk_idx)
     }
 }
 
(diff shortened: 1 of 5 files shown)


plonky2x/core/src/frontend/hash/blake2/curta.rs
```diff
@@ -379,4 +398,36 @@ mod tests {
...
+
+        builder.assert_is_equal(calculated_digest, expected_digest);
+
+        let circuit = builder.build();
+        let mut input = circuit.input();
+        input.write::<ArrayVariable<ByteVariable, MSG_LEN>>(msg_bytes[0].clone());
+        let (proof, output) = circuit.prove(&input);
+        circuit.verify(&proof, &input, &output);
+        circuit.test_default_serializers();
+    }
```
ratankaliani, line 432: Test cases should use the `blake2` crate to compute the expected hash. Also, modify test cases to be over smaller lengths (as then we can run this in CI).

plonky2x/core/src/frontend/hash/blake2/curta.rs
```diff
@@ -116,7 +116,7 @@ impl<L: PlonkParameters<D>, const D: usize> Hash<L, D, 96, true, 4> for BLAKE2B
 
     fn hash(message: Vec<u8>) -> [Self::Integer; 4] {
         let mut num_message_chunks = (message.len() as u64 / 128) + 1;
-        if num_message_chunks % 128 == 0 {
+        if message.len() % 128 == 0 && !message.is_empty() {
             num_message_chunks -= 1;
         }
```
ratankaliani, line 121: nit: (message.len() as u64 + 127) / 128 is simpler

Fine leaving it this way as well, just add a comment.

plonky2x/core/src/frontend/hash/blake2/curta.rs
```diff
@@ -227,7 +227,26 @@ impl<L: PlonkParameters<D>, const D: usize> CircuitBuilder<L, D> {
...
+        // Find out if the input length is not zero.
+        let is_zero_length = self.is_zero(input_byte_length.variable);
+        let is_not_zero_length = self.not(is_zero_length);
+
+        // Find out if we need to subtract 1 from the last chunk index.
+        let do_minus_one = self.and(is_multiple, is_not_zero_length);
+        let one = self.one();
+        let last_chunk_idx_minus_one = self.sub(last_chunk_idx, one);
+
+        self.select(do_minus_one, last_chunk_idx_minus_one, last_chunk_idx)
```
ratankaliani, line 249:
```
if input_byte_length == 0 {
      0
  } else {
      (input_byte_length - 1) / 128
  }
```

This is cleaner, doesn't require `mul` logic.

kevjue: How strongly do you feel about this?  I actually think that my implementation is more easier to understand.  E.g. the (input_byte_length - 1)/128 kind of implies the logic that I'm doing.

ratankaliani: The comments above the logic should be more clear to specify what's going on then.

I do think the following is more clear to follow: "If the message length is 0, the last idx is 0. If it is non-zero, the message length is `input_byte_length - 1` / 128. Don't need to add in the additional "find out if input length is a multiple of 128"

plonky2x/core/src/frontend/hash/blake2/curta.rs
```diff
@@ -283,13 +322,16 @@ mod tests {
...
         let bytes_length = builder.constant::<U32Variable>(msg_bytes.len() as u32);
-        let result = builder.curta_blake2b_variable(&msg.0, bytes_length);
+        let variable_result = builder.curta_blake2b_variable(&msg.0, bytes_length);
+        let fixed_result = builder.curta_blake2b(&msg.0);
 
-        let expected_digest =
-            bytes32!("7c38fc8356aa20394c7f538e3cee3f924e6d9252494c8138d1a6aabfc253118f");
+        let expected_digest = bytes32!(get_expected_digest(&msg_bytes));
+
+        bytes32!("7c38fc8356aa20394c7f538e3cee3f924e6d9252494c8138d1a6aabfc253118f");
```
ratankaliani, line 330: nit

ratankaliani: @kevjue Remove this line

plonky2x/core/src/frontend/hash/blake2/curta.rs
ratankaliani: @kevjue Please clean up the comments here ^

ratankaliani: This is personal pref, but I also try to avoid "we need to" when writing comments. Ex. "If it is ~~we need to~~ subtract 1 from the last chunk index."

plonky2x/core/src/frontend/hash/blake2/curta.rs
ratankaliani: Fixed

plonky2x/core/src/frontend/hash/blake2/curta.rs
ratankaliani: Fixed

ratankaliani (approved):
lgtm, add description to PR and feel free to merge!

PR #365, opened 2024-02-16
chore: export watch generator

dndll:
The Watch generator here is pretty useful for integrations, this just allows others to use it.

Since the module presumes the crate name in the macro we just need a use alias so that the crate and downstream dependencies can use it as it.


ratankaliani (approved):
lgtm!

PR #372, opened 2024-02-26
feat: add gateway example

ratankaliani:


contracts/script/gateway-examples/DeployAndRegisterFunction.s.sol
```diff
@@ -0,0 +1,32 @@
...
+        // SuccinctGateway address
+        address gateway = vm.envAddress("SUCCINCT_GATEWAY");
+        console.logAddress(gateway);
+
+        // Create2 salt
+        bytes32 salt = vm.envBytes32("CREATE2_SALT");
+
+        (bytes32 functionId, address verifier) =
+            SuccinctGateway(gateway).deployAndRegisterFunction(msg.sender, bytecode, salt);
+
```
mattstam, line 26: nit: returning these is generally easier to read then logging

contracts/script/gateway-examples/DeployAndRegisterFunction.s.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "forge-std/Script.sol";
+import "forge-std/Vm.sol";
+import "forge-std/console.sol";
+import {FunctionVerifier} from "./FunctionVerifier.sol";
```
mattstam, line 7: mainly concerned about where this comes from

contracts/script/gateway-examples/DeployAndRegisterFunction.s.sol
```diff
@@ -0,0 +1,32 @@
...
+
+contract DeployAndRegisterFunction is Script {
+    function run() external {
+        vm.startBroadcast();
+
+        // Get the bytecode of the FunctionVerifier contract.
+        bytes memory bytecode = type(FunctionVerifier).creationCode;
+
+        // SuccinctGateway address
+        address gateway = vm.envAddress("SUCCINCT_GATEWAY");
```
mattstam, line 18: nit: UPPER_CASE for these since we generally do that in other scripts

mattstam: `address OWNER = msg.sender` might be more clear, that way they can change this if desired

mattstam (approved):

PR #381, opened 2024-03-18
refactor: proof timeout

ratankaliani:
## Overview
- Reduce the default timeout for proofs & batch proofs.
- Let users set configuration for proof and batch proof timeout via `PROOF_BATCH_TIMEOUT_SECS` and `PROOF_TIMEOUT_SECS`.

### DIFF
--- a/plonky2x/core/src/backend/prover/remote.rs
+++ b/plonky2x/core/src/backend/prover/remote.rs
@@ -61,18 +61,25 @@ impl RemoteProver {
             .submit::<L, D>(request)
             .expect("failed to submit proof request");
 
-        // Wait for the proof to be generated.
-        const MAX_RETRIES: usize = 500;
+        // Default timeout for a proof is 60 minutes. Users can override this value by
+        // setting the PROOF_TIMEOUT_MINS environment variable.
+        const DEFAULT_PROOF_TIMEOUT_MINS: usize = 60;
+        let proof_timeout_mins =
+            env::var("PROOF_TIMEOUT_MINS").unwrap_or(DEFAULT_PROOF_TIMEOUT_MINS.to_string());
+        let proof_timeout_secs = proof_timeout_mins.parse::<u64>().unwrap() * 60;
+        let poll_delay_secs = 10;
+        let max_polls = proof_timeout_secs / poll_delay_secs;
+
         let mut status = ProofRequestStatus::Pending;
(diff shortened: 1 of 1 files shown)


plonky2x/core/src/backend/prover/remote.rs
```diff
@@ -61,18 +61,26 @@ impl RemoteProver {
             .submit::<L, D>(request)
             .expect("failed to submit proof request");
 
-        // Wait for the proof to be generated.
-        const MAX_RETRIES: usize = 500;
+        // Default timeout for a proof is 60 minutes. Users can override this value by
+        // setting the PROOF_TIMEOUT_MINS environment variable.
+        const DEFAULT_PROOF_TIMEOUT_MINS: u64 = 60;
+        let proof_timeout_mins =
+            env::var("PROOF_TIMEOUT_MINS").unwrap_or(DEFAULT_PROOF_TIMEOUT_MINS.to_string());
```
mattstam, line 68: nit: I think minutes is pretty strange as a configuration variable, bullish defining it in seconds

mattstam (approved):
small nit but otherwise lgtm

plonky2x/core/src/backend/prover/remote.rs
```diff
@@ -114,9 +122,18 @@ impl RemoteProver {
             .collect_vec();
         let (batch_id, proof_ids) = service.submit_batch(&requests)?;
 
-        const MAX_RETRIES: usize = 1500;
-        for _ in 0..MAX_RETRIES {
-            sleep(Duration::from_secs(10)).await;
+        // Default timeout for a batch proof is 60 minutes. Users can override this value by
+        // setting the BATCH_PROOF_TIMEOUT_MINS environment variable.
+        const DEFAULT_BATCH_PROOF_TIMEOUT_MINS: u64 = 60;
+        let proof_timeout_mins = env::var("BATCH_PROOF_TIMEOUT_MINS")
```
mattstam, line 128: nit: atleast in Go, it is better to read env at top-level and pass it into this function as an argument. This way you can error immediately if .env is incorrect (I guess with defaulting it might be fine)

ratankaliani: Yeah I agree if it's required, but because we have defaults, I'll keep it as this for now. As there are no default args in Rust, probably better to do it this way too.

PR #382, opened 2024-03-22
feat: network contracts

mattstam:
Changes from existing:

1. incorporate programHash - e.g function `verify(bytes32 programHash, bytes32 inputHash, bytes32 outputHash, bytes calldata proof) external returns (bool)` along with passing `programHash` through all of our other funcs
2. simplify registry - remove function ID / upgradeability and use reference to verifier address instead
    2a. maybe add transferOwnership() since doing the whitelist / custom prover calls with multisig can be annoying
3. cancun evm version - `TSTORE` saves ~20k gas on fulfillCallback and ~60k gas on fulfillCall
4. use calldata instead of memory for input / output / proof / context bytes - gas saving
5. update libs - OZ switches to custom errors which significantly reduces bytecode size
6. log verifier, callback, and prover addresses in fulfill events - helps when doing the off-chain accounting
7. separate out the admin stuff (e.g. `setFeeVault()`) into an abstract contract
8. rename contracts - NetworkGateway is NetworkStorage, NetworkAdmin, NetworkRegistry

### DIFF
--- a/contracts/foundry.toml
+++ b/contracts/foundry.toml
@@ -1,8 +1,9 @@
 [profile.default]
-solc = '0.8.16'
+solc = '0.8.25'
 src = "src"
 out = "out"
 libs = ["lib"]
+evm_version = "cancun"
 optimizer = true
 optimizer_runs = 100_000
 fs_permissions = [{ access = "read-write", path = "./"}]
--- a/contracts/src/NetworkGateway.sol
+++ b/contracts/src/NetworkGateway.sol
@@ -1,0 +1,326 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {IFeeVault} from "./payments/interfaces/IFeeVault.sol";
+import {INetworkVerifier} from "./interfaces/INetworkVerifier.sol";
+import {NetworkStorage} from "./NetworkStorage.sol";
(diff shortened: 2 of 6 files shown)


contracts/foundry.toml
```diff
@@ -1,8 +1,9 @@
 [profile.default]
-solc = '0.8.16'
+solc = '0.8.25'
 src = "src"
 out = "out"
 libs = ["lib"]
+evm_version = "cancun"
```
mattstam, line 6: cancun is live on all the chains we deploy on except scroll, and they have an ETA of early April

contracts/src/NetworkGateway.sol
```diff
@@ -0,0 +1,326 @@
...
+    error ReentrantFulfill();
+    error InvalidRequest(uint32 nonce, bytes32 expected, bytes32 actual);
+    error InvalidProof();
+    error CallbackFailed(bytes4 callbackSelector, bytes output, bytes context);
+    error CallFailed(address callback, bytes callbackData);
+    error InvalidCall(address verifier, bytes input);
+}
+
+interface INetworkGateway is INetworkGatewayEvents, INetworkGatewayErrors {
+    function requestCallback(
```
mattstam, line 61: Note: these interfaces wont be backwards compatible with olds ones since verifer/programHash is used instead of functionID

PR #385, opened 2024-04-16
feat: Make num mod public

NikolayKostadinov21:
Making the num module public enables the use of `BigUintTarget`.


taint160596 (commented):
1

